PR #31, opened 2019-08-15
Move meta classifiers

t-tte:
Added classification capability to the base-learners as the first step.

### DIFF
--- a/causalml/inference/meta/slearner.py
+++ b/causalml/inference/meta/slearner.py
@@ -43,8 +43,8 @@ class StatsmodelsOLS(object):
         return self.model.predict(X)
 
 
-class BaseSRegressor(object):
-    """A parent class for S-learner regressor classes.
+class BaseSLearner(object):
+    """A parent class for S-learner classes.
     An S-learner estimates treatment effects with one machine learning model.
     Details of S-learner are available at Kunzel et al. (2018) (https://arxiv.org/abs/1706.03461).
     """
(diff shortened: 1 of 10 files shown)


jeongyoonlee:
Thanks for taking the lead on this, @t-tte. I think it might be better to extend base meta-learner classes to base meta-classifier and meta-regressor classes similar to how scikit-learn does?

t-tte:
@jeongyoonlee - Just pushed the version with meta-classifiers and meta-regressors as separate classes.

causalml/inference/meta/slearner.py
```diff
@@ -223,3 +198,131 @@ def estimate_ate(self, X, treatment, y):
...
+                mae(y[~is_treatment], yhat_c[~is_treatment]))
+            )
+            logger.info('RMSE (Treatment): {:.6f}'.format(
+                np.sqrt(mse(y[is_treatment], yhat_t[is_treatment])))
+            )
+            logger.info(' MAE (Treatment): {:.6f}'.format(
+                mae(y[is_treatment], yhat_t[is_treatment]))
+            )
+
+        return (yhat_t - yhat_c).reshape(-1, 1)
```
yungmsh: We might want to add np.clip here to ensure values are between 0 and 1 (e.g. if yhat_c > yhat_t)

t-tte:
As discussed, leaved the BaseLearner classes as they are and extended to Regressor and Classifier subclasses per each meta-learner.

jeongyoonlee:
Thanks! Code looks great. Could you please update the tests for meta-learners with new API change? Also what about leaving `Base[S,X,T,R]Learner` in `inference.meta.__init__.py` for backward compatibility?

t-tte:
Both sound good, will push changes later today.

t-tte:
Tests have been updated.

jeongyoonlee:
Thanks @t-tte. Test code looks good as well. One fix needed: Could you please change `LRSLearner` to `LRSRegressor` in `tests/test_metrics.py` to the build pass the test (https://travis-ci.com/uber/causalml/jobs/227749855#L562)

PR #42, opened 2019-08-28
add xgboost ranking algo for rlearner regressor

yluogit:
This PR is for issue #41 
- Added xgboost with pairwise ranking objective as a derived class of BaseRRegressor 
- Fixed a typo in the `auuc_score` docstring


yluogit:
> Thanks for the PR. Can we make the `objective` and `metric` input arguments so that a user can choose between `rank:pairwise` + `auc` vs. `reg:squarederror` + `rmse` (default for regression)?

Makes sense. Will do.

yluogit:
> LGTM. Thanks!

Thanks Jeong

PR #58, opened 2019-09-23
Add feature_importance functionality for base meta learners

yungmsh:
Added a feature_importance() function for each base learner. Two methods currently implemented:

1. Gini (out-of-the-box .feature_importances_ method if underlying learner is tree-based)
2. Permutation (uses eli5's implementation of permutation importance, or mean decrease in accuracy)

Note the subtleties in which learner(s) we choose to show feature importance, for each meta-learner. (e.g. for X Learner I chose to show just the 2 tau models).


yungmsh:
Added support for shapley method, so now there's 4 different outputs you can get:

- method="gini", normalize=True
- method="gini", normalize=False
- method="permutation"
- method="shapley"

yungmsh:
Had a chat with @huigangchen offline - I will make a few more changes to this:

- Shapley values are currently implemented as np.abs(shap_values).mean(axis=0), to give an output of shape (n_features,). After the change, we will return the shapley values in its original shape (n_samples, n_features), but with some linear transformation applied for T and X Learners.
- Will add some validation for "important" feature identification by checking how the data was generated in the synthetic method(s).

yungmsh:
Discussed the latest updates with @huigangchen today. Here are our next steps:
- permutation: use predicted tau as "y"
- gini: build another tree-based model on top of the predicted tau (for all learners except for R)
- shap: update so that shap.shape[0] == X.shape[0] (instead of specific treatment vs control)

Future work
- Modify computation of shap values for x learner such that propensity score (which also depend on X) updates are also reflected in final shap values.

causalml/inference/meta/explainer.py
```diff
@@ -0,0 +1,131 @@
...
+import matplotlib.pyplot as plt
+from lightgbm import LGBMRegressor
+from copy import deepcopy
+
+VALID_METHODS = ('gini', 'permutation', 'shapley')
+
+
+class Explainer(object):
+    def __init__(self, method, control_name, X, tau, classes, model_tau=None,
+                 features=None, normalize=True, override_checks=False, r_learners=None):
```
jeongyoonlee: @yungmsh,  Could you provide docstring for the class and member functions?

causalml/inference/meta/explainer.py
```diff
@@ -0,0 +1,131 @@
...
+        self.X = X
+        self.tau = tau
+        self.classes = classes
+        self.model_tau = LGBMRegressor() if model_tau is None else model_tau
+        self.features = features
+        self.normalize = normalize
+        self.override_checks = override_checks
+        self.r_learners = r_learners
+
+        self.importance_catalog = {'gini': self.gini_importance, 'permutation': self.perm_importance}
```
jeongyoonlee: Since `importance_catalog` is not specific to a class instance created, let's move it out of the `Explainer` class.

causalml/inference/meta/explainer.py
yungmsh: Ah yes will do - docstrings are included in the individual learner scripts but I'll add them to the Explainer class as well.

PR #63, opened 2019-09-30
uplift tree visualization

zhenyuz0500:


zhenyuz0500:
Updated for code refactoring

causalml/inference/tree/plot.py
```diff
@@ -0,0 +1,168 @@
...
+Problem. 
+"""
+
+from __future__ import print_function
+from collections import defaultdict
+import numpy as np
+import pandas as pd
+import pydotplus
+
+def uplift_tree_string(decisionTree, x_names):
```
jeongyoonlee: @zhenyuz0500 Why don't we make `uplift_tree_string()` and `uplift_tree_plot()` as member functions of `DecisionTree`, and add `plot()` to `UpliftTreeClassifier` to call `fitted_uplift_tree.plot()` within.

So, at the end, users can simply call `uplift_model.plot()` to visualize the uplift tree.

causalml/inference/tree/utils.py
```diff
@@ -0,0 +1,228 @@
+"""
```
jeongyoonlee: Since these are related to feature engineering/transformation, let's move it to `causalml/features.py`.

PR #68, opened 2019-10-13
Update Quickstart guide with missing import and array indexing

paullo0106:
1. Add missing XGBRegressor import and indexing of the returning ATE estimate arrays
2. Add R Learner too and include a link to the example notebook


jeongyoonlee:
Thanks @paullo0106 for finding it out. Could you please also update `docs/examples.rst` to be consistent?

Thanks!

jeongyoonlee:
Actually, I will create a separate PR to update API docs since we have more updates to add. Thanks.

paullo0106:
> Actually, I will create a separate PR to update API docs since we have more updates to add. Thanks.

Makes sense, thanks Jeong-Yoon!

PR #79, opened 2019-11-08
Uplift tree plot contains edges connecting to wrong parent node

paullo0106:
Use the index of the node as the unique dict key rather than level+decision to avoid duplicates, so it wouldn't misuse the last node when the actual parent is another one

After fix
uplift_tree_correct copy

Before fix

uplift_tree_validate copy
(example 1)

uplift_tree_v2_d7 copy
(example 2)


paullo0106:
@zhenyuz0500 Let me know if this makes sense, thanks!

PR #80, opened 2019-11-08
Add Targeted Maximum Likelihood Estimation

jeongyoonlee:
Targeted Maximum Likelihood Estimation (TMLE) is known to be one of state-of-the-art ATE estimators. @huigangchen converted TMLE implementation in R into Python, and I'm adding it to our meta learner module.

Currently it only has `estimate_ate()` without `fit()` or `predict()` for ITE.

- add `TMLELearner()` to `inference/meta/tmle.py`.
- add `validation_with_tmle.ipynb` to `examples`.
- add `calibrate()` to `propensity.py` for propensity score calibration with GAM.


huigangchen:
It seems to me that the bias from the TMLE on X-Learner comes from the in-sample fit and ranking. If we use out-of-sample prediction the bias goes away.

jeongyoonlee:
> It seems to me that the bias from the TMLE on X-Learner comes from the in-sample fit and ranking. If we use out-of-sample prediction the bias goes away.

Thanks! I've updated the notebook with the holdout validation, and now the uplift curve with TMLE looks very close to the one with the ground truth.

causalml/inference/meta/tmle.py
```diff
@@ -0,0 +1,175 @@
...
+            else:
+                assert segment.shape[0] == X.shape[0] and segment.ndim == 1, 'Segment must be the 1-d np.array of int.'
+                segments = np.unique(segment)
+
+                _ate = []
+                _ate_lb = []
+                _ate_ub = []
+                for s in sorted(segments):
+                    logger.info('Training the TMLE learner for segment {}.'.format(s))
+                    filt = (segment == s) & (yhat_c >= 0) & (yhat_t >= 0) & (yhat_c < np.quantile(yhat_c, q=.99))
```
huigangchen: The filter on yhat_c>=0 and yhat_t>=0 may not be necessary for general situations.

jeongyoonlee: Done!

Issue #86, opened 2019-11-17
Enhance tree plot with node colors indicating positive and negative uplifts


paullo0106:
minor enhancement on [tree/plot.py ](https://github.com/uber/causalml/blob/master/causalml/inference/tree/plot.py) so it's intuitive to tell who we should target on

paullo0106:
Solved with PR #87 merge   😃

PR #87, opened 2019-11-17
Enhance tree plot with node colors indicating positive and negative uplifts (#86)

paullo0106:
For #86. Also, simplify a bit on the _totalSample_ value initiation on line 141

(sample output 1)

(sample output 2)
1116_v

### DIFF
--- a/causalml/inference/tree/plot.py
+++ b/causalml/inference/tree/plot.py
@@ -113,14 +113,12 @@ def uplift_tree_plot(decisionTree, x_names):
              ]
     i_node = 0
     dcParent = {}
-    totalSample = -1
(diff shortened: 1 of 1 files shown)


paullo0106:
@zhenyuz0500 A minor enhancement on tree visualization, kindly help me take a look when you have a chance :)

paullo0106:
The logic here:
* Light to dark blue when the uplift score is equal or greater than overall uplift (the darker the blue, the better)
* Light to dark green when the uplift score is smaller than the overall uplift (the darker the green, the worse)
* If the tree contains both training and validation set, then the validation uplift score is used

PR #97, opened 2019-12-05
accept pandas inputs for meta learners + add tests

yungmsh:
- add convert_pd_to_np in inference/meta/utils. all metalearners use this under the hood
- update synthetic summary method to accept estimators, to avoid throwing error when running pytest (default was to use causaltree, and there can be cython issues when testing locally)
- add test for accepting pandas inputs, and added one more test for TMLE


peterfoley:
would it be easier to work with if `convert_to_pd` was more like this:
```py
def convert_pd_to_np(data):
   if hasattr(pdata,"to_numpy"):
        data = data.to_numpy()
    return data
```
so that it always returns the same shape of data (vs sometimes returning tuples) and lets it work with anything that provides a `to_numpy` method?
Using to_numpy like that also lets you avoid changing a bunch of the `isinstance` calls in places where convert_pd_to_np is used. ex:
```py
if isinstance(p, np.ndarray, pd.Series):
    ...do stuff...
```
can instead be
```py
p=convert_pd_to_np(p)
if isinstance(p, np.ndarray):
    ...do stuff...
```
so that extending convert_pd_to_np in the future doesn't require changing all those lines again.

yungmsh:
Good call on suggesting`.to_numpy()` (somehow never came across it before!) - I can reduce the function to:
```python
def convert_pd_to_np(*args):
    output = [obj.to_numpy() if hasattr(obj, "to_numpy") else obj for obj in args]
    return output if len(output) > 1 else output[0]
```
The reason why I made the function able to accept an arbitrary # of args is so we can run all of the following in the same syntax:
```python
X, p, treatment, y = convert_pd_to_np(X, p, treatment, y)
X = convert_pd_to_np(X)
```

jeongyoonlee:
@yungmsh It is failing to pass the test because of missing import. Could you fix it?

PR #102, opened 2019-12-10
Xgboost objective names

### LINKED ISSUE #96: xgboost's `reg:linear` has been deprecated in favor of `reg:squarederror`
xgboost 0.90 deprecated `reg:linear` in favor of `reg:squarederror`. When you run the package tests with current xgboost, it dumps a huge number of deprecation warnings.

It looks like `reg:linear` is only mentioned by [the XGBRRegressor constructor](https://github.com/uber/causalml/blob/master/causalml/inference/meta/rlearner.py#L526), but it's a part of the existing public interface (["Effect learner objective has to be `rank:pairwise` or `reg:linear`"](https://github.com/uber/causalml/blob/master/causalml/inference/meta/rlearner.py#L549)), and `reg:linear` is required by versions before 0.90.

I'd suggest checking the xgboost version and replacing the objective with `reg:squarederror` if the version is 0.90 or greater, and tweaking the public interface to allow `effect_learner_objective='reg:squarederror'` itself and apply the same logic to pass the appropriate objective needed by the particular xgboost version.

peterfoley:
This fixes #96 by translating some xgboost objective names based on the xgboost version installed to avoid deprecation warnings. Current pip and github versions of xgboost want `reg:squarederror` in place of `reg:linear`, but `reg:squarederror` isn't compatible with old versions. 

For xgboost < 0.83, this PR translates `reg:squarederror` into `reg:linear`, and vice versa for xgboost >= 0.83 (including 0.90 and current master). When you run 
```py
>>> [XGBRRegressor(effect_learner_objective=x).effect_learner_objective 
...     for x in ['reg:linear','rank:pairwise','reg:squarederror']]
```
using xgboost==0.82, you get this:
```py
['reg:linear', 'rank:pairwise', 'reg:linear']
```
and for 0.90, you get
```py
['reg:squarederror', 'rank:pairwise', 'reg:squarederror']
```

This PR does not prevent all warnings that appear during causalml tests that use xgboost 0.90. Some of the warnings are based on a bug in 0.90 where the default constructor arguments still used `reg:linear`. Fixes for those warnings were originally implemented but created a lot of clutter for little gain, so I rolled them back in 065d018.

### DIFF
--- a/causalml/inference/meta/rlearner.py
+++ b/causalml/inference/meta/rlearner.py
@@ -10,7 +10,8 @@ from scipy.stats import norm
 from sklearn.model_selection import cross_val_predict, KFold, train_test_split
 from xgboost import XGBRegressor
 
-from causalml.inference.meta.utils import check_control_in_treatment, check_p_conditions
+from causalml.inference.meta.utils import (check_control_in_treatment, check_p_conditions,
+    clean_xgboost_objective)
 from causalml.inference.meta.explainer import Explainer
 
 logger = logging.getLogger('causalml')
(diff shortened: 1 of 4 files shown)


peterfoley:
@jeongyoonlee [The contributor instructions](https://github.com/uber/causalml/blob/master/CONTRIBUTING.md#submission) say to add the core Causal ML contributors as reviewers, but my repo permissions don't let me. Hopefully tagging counts!

jeongyoonlee:
Thanks, @peterfoley for the PR.

I already approved it, but have a second thought. What about moving `clean_dict_key()`, `metric_mapping` and `assert` for `metric_mapping` to `utils..clean_xgboost_objective()` by renaming it to `get_xgboost_objective_metric()`?

It will make code in `rlearner.py` cleaner as follows:
```python
objective, metric = get_xgboost_objective_metric(effect_learner_objective)
self.effect_learner_objective = objective
self.effect_learner_eval_metric = metric
```

PR #104, opened 2019-12-12
Make control validation consistent across all metalearners

### LINKED ISSUE #101: Add an assert to check that the control group exists
The S-Learner will run even if the default control group name ('0') is not in the treatment array, leading to seemingly paradoxical results in which eg units in the control group are predicted to have a negative treatment effect under the control condition. The other learners may have this same problem too.

To prevent the above behaviour, we should add an assert to catch situations in which the default or user-specified control group name is not in the treatment array.

peterfoley:
Fixes #101 

This tweaks the treatment/control validation logic to:

- ensure that control_name appears in the treatment vector (if control_name is not None)
- ensure that the treatment vector has more than one level

and added the check to the two metalearners that didn't have validation (S-Learner and T-Learner).

There was previously a more limited check in the R and X learners, but only if the treatment vector had two levels. This PR generalizes that and uses the same validator across all the learners.

Note that the tree methods don't have similar validation, but they might successfully throw errors if passed invalid treatment inputs. I haven't tried to address those models here.

### DIFF
--- a/causalml/inference/meta/rlearner.py
+++ b/causalml/inference/meta/rlearner.py
@@ -10,7 +10,7 @@ from scipy.stats import norm
 from sklearn.model_selection import cross_val_predict, KFold, train_test_split
 from xgboost import XGBRegressor
 
-from causalml.inference.meta.utils import check_control_in_treatment, check_p_conditions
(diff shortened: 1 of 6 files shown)


t-tte:
LGTM!

PR #105, opened 2019-12-13
Fix impurity computation

TomaszZamacinski:
Fixes the CausalTreeRegressor by correcting the impurity computation.
A DecisionTreeRegressor seeks to minimise impurity, therefore the correct value for impurity is **minus** the square of the effect. A big number is added to the new impurity to guarantee that it stays positive, otherwise non-negative min_impurity_split enforced by sklearn would prevent the tree from growing.

### DIFF
--- a/causalml/inference/tree/causaltree.pyx
+++ b/causalml/inference/tree/causaltree.pyx
@@ -88,6 +88,7 @@ cdef class CausalMSE(RegressionCriterion):
         cdef double tr_var
         cdef double ct_var
         cdef double one_over_eps = 1e5
+        cdef double big_number = 1e10
 
         for p in range(start, end):
             i = samples[p]
@@ -110,7 +111,7 @@ cdef class CausalMSE(RegressionCriterion):
         ct_var = ((node_sq_sum - node_tr_sq_sum) / node_ct -
                   (node_sum - node_tr_sum) * (node_sum - node_tr_sum) / (node_ct * node_ct))
 
-        return node_tau * node_tau - (tr_var / node_tr + ct_var / node_ct)
+        return big_number - (node_tau * node_tau - (tr_var / node_tr + ct_var / node_ct))
 
 
     cdef void children_impurity(self, double* impurity_left, double* impurity_right) nogil:
@@ -151,6 +152,7 @@ cdef class CausalMSE(RegressionCriterion):
         cdef double left_ct_var
 
         cdef double one_over_eps = 1e5
+        cdef double big_number = 1e10
 
         for p in range(start, end):
             i = samples[p]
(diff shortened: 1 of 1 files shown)


causalml/inference/tree/causaltree.pyx
```diff
@@ -110,7 +111,7 @@ cdef class CausalMSE(RegressionCriterion):
         ct_var = ((node_sq_sum - node_tr_sq_sum) / node_ct -
                   (node_sum - node_tr_sum) * (node_sum - node_tr_sum) / (node_ct * node_ct))
 
-        return node_tau * node_tau - (tr_var / node_tr + ct_var / node_ct)
+        return big_number - (node_tau * node_tau - (tr_var / node_tr + ct_var / node_ct))
```
jeongyoonlee: We don't need a constant term because it will be canceled out when calculating the improvement in impurity. Let's just flip the sign.

causalml/inference/tree/causaltree.pyx
```diff
@@ -187,8 +189,8 @@ cdef class CausalMSE(RegressionCriterion):
         left_ct_var = ((left_sq_sum - left_tr_sq_sum) / left_ct -
                         (left_sum - left_tr_sum) * (left_sum - left_tr_sum) / (left_ct * left_ct))
 
-        impurity_left[0] = left_tau * left_tau - (left_tr_var / left_tr + left_ct_var / left_ct)
-        impurity_right[0] = right_tau * right_tau - (right_tr_var / right_tr + right_ct_var / right_ct)
+        impurity_left[0] = big_number - (left_tau * left_tau - (left_tr_var / left_tr + left_ct_var / left_ct))
+        impurity_right[0] = big_number - (right_tau * right_tau - (right_tr_var / right_tr + right_ct_var / right_ct))
```
jeongyoonlee: Same as above. Let's remove the constant and just flip the sign.

TomaszZamacinski:
@jeongyoonlee on a second thought, the big_number I used might not be big enough for some applications. What can we increase it to without negative impact on numerical performance?

jeongyoonlee:
@TomaszZamacinski could you point me where non-negative `min_impurity_split` is enforced in `sklearn`? Let me dig in to it.

TomaszZamacinski:
@jeongyoonlee sklearn raises a ValueError if a user tries to set `min_impurity_split` to negative value: https://github.com/scikit-learn/scikit-learn/blob/bf24c7e3d/sklearn/tree/_classes.py#L306.
It forces nodes with impurities lower than `min_impurity_split` to be leaves: https://github.com/scikit-learn/scikit-learn/blob/bf24c7e3d6d768dddbfad3c26bb3f23bc82c0a18/sklearn/tree/_tree.pyx#L233 and: https://github.com/scikit-learn/scikit-learn/blob/bf24c7e3d6d768dddbfad3c26bb3f23bc82c0a18/sklearn/tree/_tree.pyx#L457.

peterfoley:
If negative values in `min_impurity_split` wouldn't actually break the tree building code, we could make a subclass of `DecisionTreeRegressor` with a copy/pasted version of `fit()` without that one check. I don't see any methods called within `fit()` that would let us replace `self.min_impurity_split` more cleanly before the call to `builder.build()`.

Copy/pasting 150 lines just to skip one check brings its own mess of problems, of course. Monkey-patching in subclasses of `DepthFirstTreeBuilder` and `BestFirstTreeBuilder` that do the `min_impurity_split` replacement would also work but seems even worse.

TomaszZamacinski:
The negative values in `min_impurity_split` will not break the tree building code. The parameter itself will be removed in `scikit-learn 0.25`.

Issue #54, opened 2019-09-18
Speed up UpliftRandomForestClassifier?


BrianMiner:
I am trying to fit this with only 100 trees on a dataset of roughly 200,000 records and 80 features. It takes hours to fit (and the results so far has been poor - overfit it appears, I am tuning some regularization as we speak). I am wondering if there this speed is normal and if there are any plans to improve (I did not see any option to grow the trees in parallel)? Also - on a real world data set (say from a marketing response problem) has this implementation proved to predict well?

zhenyuz0500:
Hi BrianMiner,

Thanks for the feedback!

Here are some comments and thoughts:
(1) Yes, uplift trees (KL, ED, CHI, CTS) are slow for handling large dataset for now 
(2) There is a plan to accelerate these algorithms, but it will not be implemented in the next couple months
(3) You can focus on meta learners, such as two models, x-learner, and r-learner for now, since they are relatively fast
(4) based on my experience, KL, CHI, ED are performing similarly, to save time, you can choose one of these three for training instead of all. We haven't found CTS outperforming other methods in our studies.
(5) If the features are good enough, the treatment effect is significant in ATE for the original experiment, then the uplift models would predict well - we have use cases launched in experiments that proved the models were able to provide significant results.

Hope it helps!

paullo0106:
Just some ideas on the UpliftTree optimization, we can probably start from [growDecisionTreeFrom()](https://github.com/uber/causalml/blob/master/causalml/inference/tree/models.py#L850) in the _UpliftTreeClassifier_ class:

- **Pre-sort**: sort the dataset in the beginning, and use the sorted vectors at each node evaluation (reference - _presort_ in [sklearn's BaseDecisionTree](https://github.com/scikit-learn/scikit-learn/blob/1495f6924/sklearn/tree/tree.py#L330))
- **Split**:   ([sklearn's tree splitter](https://github.com/scikit-learn/scikit-learn/blob/1495f69242646d239d89a5713982946b8ffcf9d9/sklearn/tree/_splitter.pyx#L300))
  - Do the actual data set split after finding the best candidate 
  - When the variable is numerical, instead of re-creating datasets repeatedly, use a linear algorithm that incrementally subtracts from the left and adds to the right bins to compute the new values for two bins per the splitting value.
  - Avoid evaluating again on the binary variables which were used along the route
 
These are very high-level thoughts, look forward to learning the ideas from others :)

jeongyoonlee:
[...]

PR #108, opened 2019-12-20
Speed up UpliftTreeClassifier

jeongyoonlee:
This PR is for #54.

* speeds up `UpliftTreeClassifier` by ~4x by using `numpy.array` instead of `list`.
* removes unused code from `inference.tree.models.py`

Profiling results before and after the change are as follows:

Before the change:
```bash
         4214203 function calls (4210937 primitive calls) in 2.103 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    2.094    2.094 /Users/jeong/development/causalml/causalml/inference/tree/models.py:159(fit)
      7/1    0.041    0.006    2.086    2.086 /Users/jeong/development/causalml/causalml/inference/tree/models.py:847(growDecisionTreeFrom)
     1463    0.003    0.000    0.988    0.001 /Users/jeong/development/causalml/causalml/inference/tree/models.py:501(divideSet)
     1527    0.006    0.000    0.790    0.001 /Users/jeong/development/causalml/causalml/inference/tree/models.py:782(tree_node_summary)
     1534    0.790    0.001    0.790    0.001 /Users/jeong/development/causalml/causalml/inference/tree/models.py:533(group_uniqueCounts)
     1463    0.269    0.000    0.497    0.000 /Users/jeong/development/causalml/causalml/inference/tree/models.py:529(<listcomp>)
```

After the change:
```bash
         212706 function calls (209440 primitive calls) in 0.546 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    0.539    0.539 /Users/jeong/development/causalml/causalml/inference/tree/models.py:158(fit)
      7/1    0.018    0.003    0.538    0.538 /Users/jeong/development/causalml/causalml/inference/tree/models.py:825(growDecisionTreeFrom)
     1315    0.019    0.000    0.218    0.000 /Users/jeong/development/causalml/causalml/inference/tree/models.py:759(tree_node_summary)
     1322    0.160    0.000    0.200    0.000 /Users/jeong/development/causalml/causalml/inference/tree/models.py:518(group_uniqueCounts)
      654    0.046    0.000    0.120    0.000 /Users/jeong/development/causalml/causalml/inference/tree/models.py:697(normI)
```


paullo0106:
Today I learned, this is really cool!

@jeongyoonlee I understand the optimization on `divideSet()`, `group_uniqueCounts()` and `convert_pd_to_np()` contributed to the speedup, and it makes sense to find the their execution time 
 became much more faster (**percall**). Could you shed some light on the **ncalls** decrease? Per your profiling result, the number of calls of `divideSet()`,`tree_node_summary()`, `group_uniqueCounts()` decrease significantly as well, while I don't find the relevant diffs causing this, think I might have missed something  🤔

causalml/inference/meta/tmle.py
```diff
@@ -107,13 +108,14 @@ def estimate_ate(self, X, p, treatment, y, segment=None, return_ci=False):
         Returns:
             (tuple): The ATE and its confidence interval (LB, UB) for each treatment, t and segment, s
         """
+        X, treatment, y = convert_pd_to_np(X, treatment, y)
```
paullo0106: redundant call with line 113?

causalml/inference/meta/rlearner.py
```diff
@@ -479,12 +480,13 @@ def fit(self, X, p, treatment, y, verbose=True):
             treatment (np.array or pd.Series): a treatment vector
             y (np.array or pd.Series): an outcome vector
         """
+        X, treatment, y = convert_pd_to_np(X, treatment, y)
```
paullo0106: redundant call with line 485

causalml/inference/meta/xlearner.py
```diff
@@ -250,7 +250,7 @@ def estimate_ate(self, X, p, treatment, y, bootstrap_ci=False, n_bootstraps=1000
 
         X, treatment, y = convert_pd_to_np(X, treatment, y)
         check_p_conditions(p, self.t_groups)
-        if isinstance(p, np.ndarray):
+        if isinstance(p, (np.ndarray, pd.Series)):
```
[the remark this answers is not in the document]
jeongyoonlee: Done. Thanks for catching this.

causalml/inference/meta/tmle.py
jeongyoonlee: Done. Thanks for catching this.

causalml/inference/meta/rlearner.py
```diff
@@ -586,12 +588,13 @@ def fit(self, X, p, treatment, y, verbose=True):
...
```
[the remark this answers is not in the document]
jeongyoonlee: Done. Thanks for catching this.

causalml/inference/meta/rlearner.py
jeongyoonlee: Done. Thanks for catching this.

jeongyoonlee:
> Today I learned, this is really cool!
> 
> @jeongyoonlee I understand the optimization on `divideSet()`, `group_uniqueCounts()` and `convert_pd_to_np()` contributed to the speedup, and it makes sense to find the their execution time
> became much more faster (**percall**). Could you shed some light on the **ncalls** decrease? Per your profiling result, the number of calls of `divideSet()`,`tree_node_summary()`, `group_uniqueCounts()` decrease significantly as well, while I don't find the relevant diffs causing this, think I might have missed something 🤔

Hmm, interesting. I'm not sure why. First main performance improvement was achieved by reducing two list comprehension operations against whole rows into a single masking as follows:
```python
if isinstance(value, int) or isinstance(value, float):
    splittingFunction = lambda row: row[column] >= value
else:
    splittingFunction = lambda row: row[column] == value
list1 = [row for row in rows if splittingFunction(row)]
list2 = [row for row in rows if not splittingFunction(row)]
return (list1, list2)
``` 
vs.
```python
if isinstance(value, int) or isinstance(value, float):
    filt = X[:, column] >= value
else:
    filt = X[:, column] == value

return X[filt], X[~filt], treatment[filt], treatment[~filt], y[filt], y[~filt]
...
```
Second significant improvement was from using `tuple` instead of `dict` in `group_uniqueCounts()`. When we have only `0` and `1` as keys, we don't need to use `dict` which add significant overhead.

Issue #107, opened 2019-12-17
Add check to ensure that propensity scores are not ones and zeros only


t-tte:
If you pass a propensity score that contains only ones and zeros, the X-learner fails with the following ambiguous error:

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-186-7587b62cdf36> in <module>
     56     ps = ps_model.predict(X[test])
     57 
---> 58     cate_pred = uplift_model.predict(X[test], ps)
     59 
     60     # Get recommended treatment group and actual value diff

~/phoenix-worker/environments/python3/lib/python3.6/site-packages/causalml/inference/meta/xlearner.py in predict(self, X, p, treatment, y, return_components, verbose)
    594             dhat_ts[group] = model_tau_t.predict(X)
    595 
--> 596             _te = (p[group] * dhat_cs[group] + (1 - p[group]) * dhat_ts[group]).reshape(-1, 1)
    597             te[:, i] = np.ravel(_te)
    598 

TypeError: can't multiply sequence by non-int of type 'float'
```

It may be useful to add a check that the propensity score contains more than two unique values. In particular, this prevents the user from mistakenly passing binary labels as the propensity score.

t-tte:
Addressed in PR #110

PR #110, opened 2019-12-20
Add sanity checks for propensity score values

t-tte:
Addresses #107

### DIFF
--- a/causalml/inference/meta/utils.py
+++ b/causalml/inference/meta/utils.py
@@ -31,7 +31,8 @@ def check_p_conditions(p, t_groups):
 
     if isinstance(p, dict):
         for t_name in t_groups:
-            assert (0 + eps < p[t_name]).all() and (p[t_name] < 1 + eps).all(), 'The values of p should lie within the (0, 1) interval.'
+            assert (0 + eps < p[t_name]).all() and (p[t_name] < 1 + eps).all(), \
+                'The values of p should lie within the (0, 1) interval.'
 
 
 def check_explain_conditions(method, models, X=None, treatment=None, y=None):
--- a/causalml/inference/meta/utils.py
+++ b/causalml/inference/meta/utils.py
@@ -25,6 +25,17 @@ def check_p_conditions(p, t_groups):
     if isinstance(p, (np.ndarray, pd.Series)):
         assert t_groups.shape[0] == 1, \
             'If p is passed as an np.ndarray, there must be only 1 unique non-control group in the treatment vector.'
+        assert len(np.unique(p)) > 2, \
+            'p should have more than two distinct values'
+        assert (0 < p).all() and (p < 1).all(), \
+            'The values of p should lie within the (0, 1) interval.'
+            
+    if isinstance(p, dict):
+        for t_name in t_groups:
+            assert len(np.unique(p[t_name])) > 2, \
+                'p should have more than two distinct values'
+            assert (0 < p[t_name]).all() and (p[t_name] < 1).all(), \
+                'The values of p should lie within the (0, 1) interval.'
 
 
 def check_explain_conditions(method, models, X=None, treatment=None, y=None):
--- a/causalml/inference/meta/utils.py
+++ b/causalml/inference/meta/utils.py
@@ -20,22 +20,18 @@ def check_treatment_vector(treatment, control_name=None):
 
 
 def check_p_conditions(p, t_groups):
+    eps = np.finfo(float).eps
     assert isinstance(p, (np.ndarray, pd.Series, dict)), \
         'p must be an np.ndarray, pd.Series, or dict type'
     if isinstance(p, (np.ndarray, pd.Series)):
         assert t_groups.shape[0] == 1, \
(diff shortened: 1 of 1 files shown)


jeongyoonlee:
We can ignore failing tests due to missing imports from the previous merge. I fixed it in #108.

ppstacy:
@t-tte thanks for adding the check here! I have a question regarding at least two distinct values for the propensity score. Recently when we use X/R-learner for the Bidder project, we used constant value (0.5) because of the equal chance for intervention in both the control and treatment group. So it probably will fail here.  Any suggestions for how should we handle this? 

cc @jeongyoonlee  who probably has more context here as well.

t-tte:
Hi @ppstacy - that's a fair comment given that using a constant instead of a propensity score is suggested as an option in Kunzel et al (2019). That said, I think it would make sense to keep the propensity scores (and the associated checks) as one thing, and the option to pass a constant to the X-Learner as another thing. For example, perhaps we can add an optional argument to the X-Learner that lets the user fit the model with a constant instead of a propensity score.

causalml/inference/meta/utils.py
```diff
@@ -25,6 +25,17 @@ def check_p_conditions(p, t_groups):
...
             'If p is passed as an np.ndarray, there must be only 1 unique non-control group in the treatment vector.'
+        assert len(np.unique(p)) > 2, \
+            'p should have more than two distinct values'
+        assert (0 < p).all() and (p < 1).all(), \
+            'The values of p should lie within the (0, 1) interval.'
+            
+    if isinstance(p, dict):
+        for t_name in t_groups:
+            assert len(np.unique(p[t_name])) > 2, \
+                'p should have more than two distinct values'
```
jeongyoonlee: As @ppstacy pointed out, we may use a constant propensity e.g. in RCT. I don't think we need this check even for R-learner as long as propensity values are not extreme (i.e. close to either 0 or 1).

causalml/inference/meta/utils.py
```diff
@@ -25,6 +25,17 @@ def check_p_conditions(p, t_groups):
     if isinstance(p, (np.ndarray, pd.Series)):
         assert t_groups.shape[0] == 1, \
             'If p is passed as an np.ndarray, there must be only 1 unique non-control group in the treatment vector.'
+        assert len(np.unique(p)) > 2, \
+            'p should have more than two distinct values'
+        assert (0 < p).all() and (p < 1).all(), \
+            'The values of p should lie within the (0, 1) interval.'
```
jeongyoonlee: Should we consider using `0 + eps` and `1 - eps` for stability?

causalml/inference/meta/utils.py
```diff
@@ -25,6 +25,17 @@ def check_p_conditions(p, t_groups):
...
+            
+    if isinstance(p, dict):
+        for t_name in t_groups:
+            assert len(np.unique(p[t_name])) > 2, \
+                'p should have more than two distinct values'
+            assert (0 < p[t_name]).all() and (p[t_name] < 1).all(), \
+                'The values of p should lie within the (0, 1) interval.'
```
jeongyoonlee: Same as in line 30. Let's use `0 + eps` and `1 - eps` instead.

peterfoley:
As a user of many tools that have tried to protect me from myself, I don’t like requiring >2 unique values. If for whatever reason a user has a propensity score with only 2 unique values, shouldn’t they still be allowed to use the package?

If folks are passing categorical values by mistake, they’ll probably have integer values, and the second assertion will catch those cases. Since propensity scores in most causal inference applications have to be in (0,1), that assertion matches user expectations.

causalml/inference/meta/utils.py
```diff
@@ -20,11 +20,19 @@ def check_treatment_vector(treatment, control_name=None):
 
 
 def check_p_conditions(p, t_groups):
+    eps = np.finfo(float).eps
     assert isinstance(p, (np.ndarray, pd.Series, dict)), \
         'p must be an np.ndarray, pd.Series, or dict type'
     if isinstance(p, (np.ndarray, pd.Series)):
         assert t_groups.shape[0] == 1, \
             'If p is passed as an np.ndarray, there must be only 1 unique non-control group in the treatment vector.'
+        assert (0 + eps < p).all() and (p < 1 + eps).all(), \
```
jeongyoonlee: `p < 1 - eps` instead of `p < 1 + eps`.

causalml/inference/meta/utils.py
```diff
@@ -20,11 +20,19 @@ def check_treatment_vector(treatment, control_name=None):
...
+            'The values of p should lie within the (0, 1) interval.'
+
+    if isinstance(p, dict):
+        for t_name in t_groups:
+            assert (0 + eps < p[t_name]).all() and (p[t_name] < 1 + eps).all(), \
```
jeongyoonlee: Same as above.

jeongyoonlee:
LGTM. Thanks for fixing this over the weekend. Please try not to work too hard. :p

Issue #70, opened 2019-10-16
Propensity score requirement in X-learner and R-learner


nsalas24:
First, thanks for open-sourcing this package. I've learned a lot!

I'm wondering if there's a particular reason why the user must pass self generated propensity scores to be used in the X-learner and R-learner? While it most likely forces the user to understand how well calibrated the scores are, I would think there's validity in estimating them 'under the hood' using the `learner` or another supplied model parameterization (similar to how `BaseRLearner.outcome_learner` can be optionally specified). 

In terms of the X-learner, while Kunzel et al. state they have good experiences specifying g(x) as the propensity scores, its worth noting g(x) is simply a weighting function that is chosen to minimize the variance of the CATEs. Stating this in the documentation / naming conventions might be helpful also?

Thanks

t-tte:
Thanks for the comment. It may be a good idea to implement some optional "under the hood" calculation of the propensity score in future versions. The package comes with the simple ElasticNetPropensityModel class for urgent needs.

As you suggest, in X/R-Learner the propensity score doesn't play the same role as it does in observational causal inference. There is a case for highlighting this in the documentation as we've found the terminology of propensity scores can cause confusion for users.

nsalas24:
@t-tte I can take a first stab at this.

swager:
I came across this thread serendipitously, and just wanted to add a quick comment.

The way the X-learner and the R-learner use the propensity score is very different. The X-learner in fact just uses the propensity score as a weighting function for variance minimization, and so having accurate propensity estimates is not so important.

In contrast, the R-learner relies crucially on the propensity score for accurate CATE estimation. The R-learner is motivated by an orthogonal estimating equation for the CATE, and so the propensity score plays the same role in the R-learner as it does in many other observational study methods (including the work on double machine learning from Chernozhukov & co-authors). Having reasonably accurate propensity score estimates really helps the R-learner a lot.

The upshot is that if you want an interface that doesn't require the user to prove a propensity score for the R-learner, then you need to add functionality that learns an (accurate, cross-fitted) propensity score behind the scenes.

Finally, thanks to the maintainers of this package for all your work on this!

jeongyoonlee:
Thanks, @swager for your comment. It means a lot to the team to have kind words from you, the author of the approaches we implemented!

As suggested, we decided to simply the APIs for X and R learners to be consistent with other learners by making the propensity score input optional. If the propensity score is not provided, we will predict it under the hood.

Thanks again!

[...]

PR #113, opened 2019-12-27
make propensity score optional in X and R Learner

ppstacy:
This PR is for #70 

- Make propensity score optional in X and R Learner; If propensity score didn't provide, will run  a simple `ElasticNetPropensityModel()` for X Learner and cross-fitted for R Learner
- Update the example notebook for the cases without propensity score input


causalml/inference/meta/rlearner.py
```diff
@@ -66,21 +67,37 @@ def __repr__(self):
                                          self.model_mu.__repr__(),
                                          self.model_tau.__repr__()))
 
-    def fit(self, X, p, treatment, y, verbose=True):
+    def fit(self, X, treatment, y, p=None, return_p_score=False, verbose=True):
```
jeongyoonlee: Instead of returning `p` in `fit()`, why don't we make it available as a member variable and member function e.g. `self.propensity` and `self.get_propensity()`? We can initialize `self.propensity` as `None` and set it in `fit()` with either the input argument `p` if passed or the one computed internally here.

causalml/inference/meta/rlearner.py
```diff
@@ -271,6 +304,63 @@ def bootstrap(self, X, p, treatment, y, size=10000):
         te_b = self.predict(X=X)
         return te_b
 
+    def run_propensity_score(self, X, treatment, X_pred=None, treatment_pred=None, cv=None,
```
jeongyoonlee: The name of `run_propensity_score` is not straightforward to me. We don't "run" the propensity score, but either "compute", "calculate", or "predict" it.

causalml/inference/meta/rlearner.py
```diff
@@ -271,6 +304,63 @@ def bootstrap(self, X, p, treatment, y, size=10000):
...
+                logger.info('Training a propensity model for CV #{}'.format(i_fold))
+                p_model.fit(X[i_trn], treatment[i_trn])
+                model_dict[i_fold] = p_model
+                if X_pred is None:
+                    p[i_val] = p_model.predict(X[i_val])
+                else:
+                    p_fold = np.zeros_like(treatment_pred, dtype=float)
+                    p_fold += p_model.predict(X_pred)
+                    n_fold += 1
+            p = p_fold/n_fold
```
jeongyoonlee: To get the number of folds, we can use `cv. get_n_splits(X)` without using `n_fold`.

causalml/inference/meta/rlearner.py
```diff
@@ -271,6 +304,63 @@ def bootstrap(self, X, p, treatment, y, size=10000):
         te_b = self.predict(X=X)
         return te_b
 
+    def run_propensity_score(self, X, treatment, X_pred=None, treatment_pred=None, cv=None,
+                             calibrate_propensity=True, return_model=False):
```
jeongyoonlee: `calibrate_propensity` --> `calibrate`. It's obvious that we're calibrating propensity scores.

causalml/inference/meta/rlearner.py
```diff
@@ -271,6 +304,63 @@ def bootstrap(self, X, p, treatment, y, size=10000):
...
+            logger.info('Calibrating propensity scores.')
+            if X_pred is None:
+                p = calibrate(p, treatment)
+            else:
+                p = calibrate(p, treatment_pred)
+
+        if return_model:
+            return p, model_dict
+        else:
+            return p
```
jeongyoonlee: It's not a good practice to allow different return types and/or different numbers of return variables. Instead, we can set a member variable, e.g. `self.model_propensity`, which can be either a single model object or a list of `n_fold` models.

causalml/inference/meta/rlearner.py
ppstacy: Thanks for the suggestion here! Changed accordingly.

causalml/inference/meta/rlearner.py
ppstacy: Yeah, it makes sense! Changed accordingly.

causalml/inference/meta/rlearner.py
ppstacy: Thanks for the suggestion here! Changed accordingly.

causalml/inference/meta/rlearner.py
ppstacy: Thanks for the suggestion here! Changed accordingly.

causalml/inference/meta/rlearner.py
ppstacy: I changed to `calibrate_p` to be shorter because we import the same function here  `from causalml.propensity import calibrate`

causalml/inference/meta/rlearner.py
```diff
@@ -271,6 +299,56 @@ def bootstrap(self, X, p, treatment, y, size=10000):
         te_b = self.predict(X=X)
         return te_b
 
+    def compute_propensity_score(self, X, treatment, X_pred=None, treatment_pred=None, cv=None, calibrate_p=True):
```
jeongyoonlee: This function doesn't access anything in the class instance. Let's move it to `propensity.py` and import from it. It will also prevent from repeating this code in `xlearner.py` as well.

causalml/inference/meta/rlearner.py
```diff
@@ -271,6 +299,56 @@ def bootstrap(self, X, p, treatment, y, size=10000):
...
+
+            if X_pred is not None:
+                p = p_fold/cv.get_n_splits()
+        else:
+            p_model.fit(X, treatment)
+            p_model_dict['all training'] = p_model
+            if X_pred is None:
+                p = p_model.predict(X_pred)
+            else:
+                p = p_model.predict(X_pred)
```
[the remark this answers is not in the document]
ppstacy: Thanks for catching this!

causalml/inference/meta/rlearner.py
ppstacy: Made the change. Yeah, this is a great suggestion and will make the code simpler. Thanks!

causalml/inference/meta/xlearner.py
```diff
@@ -570,21 +606,42 @@ def fit(self, X, treatment, y):
...
+                treatment_filt = treatment[mask]
+                X_filt = X[mask]
+                w_filt = (treatment_filt == group).astype(int)
+                w = (treatment == group).astype(int)
+                p[group], p_model[group] = compute_propensity_score(X=X_filt, treatment=w_filt,
+                                                                    X_pred=X, treatment_pred=w)
+            self.propensity_model = p_model
+            self.propensity = p
+        else:
+            check_p_conditions(p, self.t_groups)
```
jeongyoonlee: I think that we should not train a propensity model in `predict()` but use either propensity scores provided or propensity model(s) trained during `fit()` instead.

Then, we can make sure that we're using the same propensity model(s) for `fit()` and `predict()`. Also we don't need to return propensity scores in `predict()` as well.

ppstacy: Hi @jeongyoonlee, thanks for reviewing it! For X-Learner, we didn't need `propensity_score`  in `fit()` and only needed when calculating the predictions of treatment effects which we have in `predict()` step.  Our original `fit()` function didn't require `p`, so I didn't add there.  For R-Learner, we did what you suggested here train a propensity model in `fit()` and return the propensity score and models(s).

jeongyoonlee: I see. Then, let's set `self.propensity` and `self.propensity_model` as in R learner without returning `p` in `predict()`.

ppstacy: Thanks! Changed accordingly.

ppstacy:
this commit for 1. in `compute_propensity_score()` when run cross validation, prevent running prediction if the data in training; 2. For x learner, in predict() use trained model to generate propensity scores instead of directly calling self.propensity

jeongyoonlee:
@ppstacy Tests are failing. Could you look into the errors from the tests?

causalml/inference/meta/xlearner.py
```diff
@@ -618,7 +712,11 @@ def predict(self, X, p, treatment=None, y=None, return_components=False, verbose
...
-        if not return_components:
+        if (not return_components) and (not return_p_score):
             return te
-        else:
+        elif (not return_components) and (return_p_score):
+            return te, p
+        elif (return_components) and (not return_p_score):
             return te, dhat_cs, dhat_ts
+        else:
+            return te, dhat_cs, dhat_ts, p
```
jeongyoonlee: Do we need to return `p` here?

causalml/inference/meta/xlearner.py
```diff
@@ -570,21 +655,30 @@ def fit(self, X, treatment, y):
...
+            return_p_score (bool, optional): whether to return propensity score
+            verbose (bool, optional): whether to output progress logs
         Returns:
             (numpy.ndarray): Predictions of treatment effects.
         """
         X, treatment, y = convert_pd_to_np(X, treatment, y)
-        check_p_conditions(p, self.t_groups)
+
+        if p is None:
+            p = self.propensity
```
jeongyoonlee: As discussed, we need to generate `p` using `self.propensity_model` trained during `fit()` instead of using `self.propensity` because `X` in `predict()` might not be the same in `X` in `fit()`.

causalml/inference/meta/xlearner.py
```diff
@@ -229,33 +282,37 @@ def fit_predict(self, X, p, treatment, y, return_ci=False, n_bootstraps=1000, bo
...
             n_bootstraps (int): number of bootstrap iterations
             bootstrap_size (int): number of samples per bootstrap
-            verbose (str): whether to output progress logs
-
         Returns:
             The mean and confidence interval (LB, UB) of the ATE estimate.
         """
-        te, dhat_cs, dhat_ts = self.fit_predict(X, p, treatment, y, return_components=True)
-
+        self.fit(X, treatment, y, p)
```
jeongyoonlee: Why do we replace `self.fit_predict()` with separate calls for `self.fit()` and `self.predict()`?

causalml/inference/meta/rlearner.py
```diff
@@ -267,7 +297,7 @@ def bootstrap(self, X, p, treatment, y, size=10000):
         p_b = {group: _p[idxs] for group, _p in p.items()}
```
jeongyoonlee: Let's change the order of input arguments in `bootstrap()` to be the same as `fit()`. i.e. `bootstrap(self, X, treatment, y, p, size)`

causalml/inference/meta/xlearner.py
```diff
@@ -319,9 +374,10 @@ def bootstrap(self, X, p, treatment, y, size=10000):
         """Runs a single bootstrap. Fits on bootstrapped sample, then predicts on whole population."""
```
jeongyoonlee: Same as above. Let's reorder the input arguments in `bootstrap()`.

PR #115, opened 2019-12-31
Causaltree refactor

### LINKED ISSUE #106: Remove the need for big_number in causaltree.pyx
Creating an issue to organize discussion that started in #105 (@jeongyoonlee, @TomaszZamacinski).

Summary of the issue is that sklearn's `DecisionTreeRegressor.fit()` raises` ValueError` if `min_impurity_split<0` and the `TreeBuilder`'s stop if `impurity<min_impurity_split`. Causal tree impurities can be negative, so sklearn's restriction doesn't help and the current workaround adds a large constant `big_number` to try to keep `impurity>0`.

`big_number` has to be large enough to ensure `impurity>0` in a large number of cases, but large values of `big_number` reduce numeric precision. It would be preferable to dodge sklearn's checks so that we can set `min_impurity_split` to negative infinity and ensure it has no impact on the tree construction.

While the `min_impurity_split` argument to `DecisionTreeRegressor` is deprecated and will be removed, I'm not sure whether it will be removed from the `TreeBuilder` classes, or whether it will just be fixed at 0 by `DecisionTreeRegressor` and not accessible to users.

peterfoley:
Fixes #106. Replicates a subset of the checks from [sklearn's BaseDecisionTree fit method](https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/tree/_classes.py#L140).

I'd like a second set of eyes to look over the validation checks that I've included/excluded. Most of the excluded ones are either specifically not desired, or they're related to parameters that had previously used default values of `DecisionTreeRegressor`'s constructor and are now hardcoded. 

It passes tests, but I haven't dug into performance vs master yet.

### DIFF
--- a/causalml/inference/tree/causaltree.pyx
+++ b/causalml/inference/tree/causaltree.pyx
@@ -278,6 +278,8 @@ class CausalTreeRegressor(object):
                                  "or in (0, 0.5], got %s"
                                  % self.min_samples_leaf)
             min_samples_leaf = int(ceil(self.min_samples_leaf * n_samples))
+        max_depth = (np.iinfo(np.int32).max if self.max_depth is None
+                     else self.max_depth)
 
         self.tree = Tree(
             n_features = n_features,
@@ -288,17 +290,17 @@ class CausalTreeRegressor(object):
             n_outputs = n_outputs)
         splitter = BestSplitter(criterion = CausalMSE(1, X.shape[0]),
             max_features = n_features,
-            min_samples_leaf = self.min_samples_leaf,
+            min_samples_leaf = min_samples_leaf,
             min_weight_leaf = 0, # from DecisionTreeRegressor default
-            random_state = self.random_state)
+            random_state = random_state)
         # hardcoded values below come from defaults values in
         #   sklearn.tree._classes.DecisionTreeRegressor
         builder = DepthFirstTreeBuilder(
             splitter = splitter,
             min_samples_split = 2,
-            min_samples_leaf = self.min_samples_leaf,
+            min_samples_leaf = min_samples_leaf,
             min_weight_leaf = 0,
-            max_depth = self.max_depth,
+            max_depth = max_depth,
             min_impurity_decrease = 0,
             min_impurity_split = float("-inf"))
         builder.build(
@@ -318,6 +320,7 @@ class CausalTreeRegressor(object):
         Returns:
             (numpy.ndarray): Predictions of treatment effects.
         """
+        X = check_array(X, dtype=DTYPE, accept_sparse="csr")
         return self.tree.predict(X).reshape((-1, 1))
 
     def fit_predict(self, X, treatment, y, return_ci=False, n_bootstraps=1000, bootstrap_size=10000, verbose=False):
--- a/causalml/inference/tree/causaltree.pyx
+++ b/causalml/inference/tree/causaltree.pyx
@@ -96,7 +96,6 @@ cdef class CausalMSE(RegressionCriterion):
         cdef double tr_var
         cdef double ct_var
         cdef double one_over_eps = 1e5
-        cdef double big_number = 1e10
 
         for p in range(start, end):
             i = samples[p]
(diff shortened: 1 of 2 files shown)


causalml/inference/tree/causaltree.pyx
```diff
@@ -234,19 +242,72 @@ class CausalTreeRegressor(object):
...
+        # hardcoded values below come from defaults values in
+        #   sklearn.tree._classes.DecisionTreeRegressor
+        builder = DepthFirstTreeBuilder(
+            splitter = splitter,
+            min_samples_split = 2,
+            min_samples_leaf = self.min_samples_leaf,
+            min_weight_leaf = 0,
+            max_depth = self.max_depth,
+            min_impurity_decrease = 0,
+            min_impurity_split = float("-inf"))
```
jeongyoonlee: Thanks @peterfoley! I ran the [test notebook](https://gist.github.com/jeongyoonlee/422312757f8c01e825a54b7f71de1ae6) and confirmed that it works the same as before the change. 

Could you please remove `big_number` in `node_impurity()` and `children_impurity()` as well?

jeongyoonlee: Oops. I ran it with your `master` instead of this branch. With this branch, the notebook raises an error at Cell 6 as follows:
Code:
```python
# Causal Tree
learner_ct = CausalTreeRegressor(min_samples_leaf=500, random_state=42)
learner_ct.fit(X=X_trn, treatment=w_trn, y=y_trn)
cate_ct = learner_ct.predict(X=X_tst).flatten()
```
Error:
```bash
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-6-d21a54e6e454> in <module>
      1 # Causal Tree
      2 learner_ct = CausalTreeRegressor(min_samples_leaf=500, random_state=42)
----> 3 learner_ct.fit(X=X_trn, treatment=w_trn, y=y_trn)
      4 cate_ct = learner_ct.predict(X=X_tst).flatten()

causalml/inference/tree/causaltree.pyx in causalml.inference.tree.causaltree.CausalTreeRegressor.fit()

sklearn/tree/_splitter.pyx in sklearn.tree._splitter.Splitter.__cinit__()

TypeError: __cinit__() takes exactly 6 positional arguments (5 given)
```

jeongyoonlee: The error above was caused with `sklearn==0.21`, and after upgrading it to `sklearn==0.22`, I got a new error:
```bash
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-7-d21a54e6e454> in <module>
      1 # Causal Tree
      2 learner_ct = CausalTreeRegressor(min_samples_leaf=500, random_state=42)
----> 3 learner_ct.fit(X=X_trn, treatment=w_trn, y=y_trn)
      4 cate_ct = learner_ct.predict(X=X_tst).flatten()

causalml/inference/tree/causaltree.pyx in causalml.inference.tree.causaltree.CausalTreeRegressor.fit()

sklearn/tree/_tree.pyx in sklearn.tree._tree.DepthFirstTreeBuilder.__cinit__()

TypeError: an integer is required
```

causalml/inference/tree/causaltree.pyx
```diff
@@ -259,7 +320,7 @@ class CausalTreeRegressor(object):
         Returns:
             (numpy.ndarray): Predictions of treatment effects.
         """
-        return self.model.predict(X).reshape((-1, 1))
```
jeongyoonlee: Need to convert the type of `X`:
```python
X = check_array(X, dtype=DTYPE, accept_sparse="csr")
```

causalml/inference/tree/causaltree.pyx
jeongyoonlee: I added a couple of comments for changes as below. After those changes, it now runs the notebook fine and produces the compatible results.

Could you please add those fixes? Thanks!

peterfoley:
thanks @jeongyoonlee (especially for catching the fact that I forgot to remove big_number! :man_facepalming:)

How would you like me to address sklearn 0.21? For the time being, I've kicked up the required version in requirements.txt, but I could instead do version checks like xgboost needed for objective name selection.

jeongyoonlee:
LGTM. For now, let's keep it as is with a minimal change in `requirements.txt`. Thanks and happy new year!

PR #121, opened 2020-01-07
explainer.py small fix and notebook update

yluogit:
explainer.py small fix: 
`if self.tau.ndim == 1` will return error when self.tau is None. It happens when calling `slearner.plot_shap_values(shap_dict=shap_slearner)` where shap_dict is given and thus no need to provide tau. 

notebook update: 
added comparison between current perm_importance implementation and sklearn's implementation. results are similar but sklearn implementation is >2x faster since no cross-validation is used


yungmsh:
LGTM. The importances look basically the same when comparing eli5 and sklearn, so that's good. I remember you mentioned sklearn's implementation is faster, but I don't see time recordings for eli5. (I do remember eli5 was a little slow when I ran it but wondering if you had printed the times as well). 

S/T/X are consistent, the only exception is R learner, because the way the importance is calculated under the hood is actually subtly different:
- https://github.com/uber/causalml/blob/yl-explainer-update/causalml/inference/meta/explainer.py#L127-L133: if self.r_learners is provided, it uses the R learner itself as the estimator (as opposed to building a new estimator on top, like S/T/X do).
- If we wanted to change that behavior, we can pass None into the r_learners parameter here: https://github.com/uber/causalml/blob/yl-explainer-update/causalml/inference/meta/rlearner.py#L296
- We can discuss this w/ the group, but I feel like it might be better (for the sake of consistency) to pass None into the r_learners parameter as default behavior.

yluogit:
> LGTM. The importances look basically the same when comparing eli5 and sklearn, so that's good. I remember you mentioned sklearn's implementation is faster, but I don't see time recordings for eli5. (I do remember eli5 was a little slow when I ran it but wondering if you had printed the times as well).
> 
> S/T/X are consistent, the only exception is R learner, because the way the importance is calculated under the hood is actually subtly different:
> 
> * https://github.com/uber/causalml/blob/yl-explainer-update/causalml/inference/meta/explainer.py#L127-L133: if self.r_learners is provided, it uses the R learner itself as the estimator (as opposed to building a new estimator on top, like S/T/X do).
> * If we wanted to change that behavior, we can pass None into the r_learners parameter here: https://github.com/uber/causalml/blob/yl-explainer-update/causalml/inference/meta/rlearner.py#L296
> * We can discuss this w/ the group, but I feel like it might be better (for the sake of consistency) to pass None into the r_learners parameter as default behavior.

I did time the current eli5 implementation in the S Learner section of the notebook: you'll find the current implementation took 3s and sklearn implementation took 0.9s. 

Good note on the R Learner. I knew you implemented the option to pass a pre-fit R Learner into the explainer but didn't realize it was the default behavior when calling `rlearner.get_importance()` so thanks for pointing that out. I agree that consistency is important when users plug in different meta leaners into the same `get_importance()` function so I would suggest default the r_learners parameter to None or even remove that parameter at all in case people find it confusing. I don't have a strong opinion on it though.

yungmsh:
@yluogit ah yes I missed the eli5 in s learner part. Yeah why don't we go ahead and update the default method in R learner case to still build a new model on top of tau predictions like S/T/X.

yluogit:
> @yluogit ah yes I missed the eli5 in s learner part. Yeah why don't we go ahead and update the default method in R learner case to still build a new model on top of tau predictions like S/T/X.

sounds good. will do that and replace eli5 with sklearn's permutation importance in the next PR

Issue #122, opened 2020-01-09
Replace ELI5 permutation importance with sklearn's 


yluogit:
**Is your feature request related to a problem? Please describe.**
Sklearn released `sklearn.inspection.permutation_importance` in version 0.22. If we use it instead of ELI5 we can remove a dependency 

**Describe the solution you'd like**
Implement sklearn's permutation importance in `explainer.py`

**Describe alternatives you've considered**

**Additional context**
https://scikit-learn.org/stable/modules/generated/sklearn.inspection.permutation_importance.html#sklearn.inspection.permutation_importance

PR #123, opened 2020-01-09
Replace ELI5 permutation importance with sklearn's

yluogit:
This PR is for #122
* replaced eli5 permutation importance with `sklearn.inspection.permutation_importance` in `explainer.py`, resulting in >4x improvement in speed on the synthetic data (0.7s vs. 3s)
* made corresponding changes in S/T/X/R Learners 
* changed default behavior of `BaseRLearner.get_importance()` to not input a pre-fit estimator, to be consistent with other meta learners
* removed eli5 from requirements.txt


yluogit:
> LGTM!
> 
> (cc @jeongyoonlee) sklearn only goes up to 0.20.4 for python3.4 or lower, so this feature won't be available to users below python3.5 in the new release. Also just validated, on a python2.7 environment, running `pip install --upgrade causalml` only gets to 0.4.0.

thank you!

PR #126, opened 2020-01-13
change elasticnetcv to logisticregressioncv

yungmsh:
Propensity model was using ElasticNetCV previously (which is a regression implementation). Changed this to LogisticRegressionCV. Explicitly imported and used StratifiedKFold (setting shuffle=True) because the default cv sets shuffle=False which can be potentially problematic.

### DIFF
--- a/causalml/propensity.py
+++ b/causalml/propensity.py
@@ -15,20 +15,22 @@ class ElasticNetPropensityModel(object):
         model (sklearn.linear_model.ElasticNetCV): a propensity model object
     """
 
-    def __init__(self, n_fold=5, clip_bounds=(1e-3, 1 - 1e-3), random_state=None):
+    def __init__(self, n_fold=5, clip_bounds=(1e-3, 1 - 1e-3), l1_ratios=np.linspace(0,1,5), random_state=None):
         """Initialize a propensity model object.
 
         Args:
             n_fold (int): the number of cross-validation fold
             clip_bounds (tuple): lower and upper bounds for clipping propensity scores. Bounds should be implemented
                 such that: 0 < lower < upper < 1, to avoid division by zero in BaseRLearner.fit_predict() step.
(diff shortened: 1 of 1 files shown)


causalml/propensity.py
```diff
@@ -15,19 +15,22 @@ class ElasticNetPropensityModel(object):
...
             clip_bounds (tuple): lower and upper bounds for clipping propensity scores. Bounds should be implemented
                 such that: 0 < lower < upper < 1, to avoid division by zero in BaseRLearner.fit_predict() step.
+            l1_ratios (array-like): array of l1 ratios (0 <= value <= 1) to iterate through in CV estimator
             random_state (numpy.random.RandomState or int): RandomState or an int seed
 
         Returns:
             None
         """
-        self.model = ElasticNetCV(cv=n_fold, random_state=random_state)
+        cv = StratifiedKFold(n_splits=n_fold, shuffle=True, random_state=random_state)
```
jeongyoonlee: Let's make `cv` accessible as `self.cv` so that in case that users can use the same `cv` across the propensity and treatment effect models (e.g. R learner).

causalml/propensity.py
```diff
@@ -15,19 +15,22 @@ class ElasticNetPropensityModel(object):
         model (sklearn.linear_model.ElasticNetCV): a propensity model object
     """
 
-    def __init__(self, n_fold=5, clip_bounds=(1e-3, 1 - 1e-3), random_state=None):
+    def __init__(self, n_fold=5, clip_bounds=(1e-3, 1 - 1e-3), l1_ratios=np.linspace(0,1,5), random_state=None):
```
jeongyoonlee: Let's allow users to pass `cv` as an input argument. We can set it to `None` by default and if not given, we can generate it as in line 31.

causalml/propensity.py
yungmsh: Done!

causalml/propensity.py
yungmsh: Done!

PR #133, opened 2020-02-12
Add synthetic dataset from CEVAE paper

### LINKED ISSUE #120: Add CEVAE synthetic data generation process
As a start of  the migration/add a wrapper of  `CEVAE` under `CausalML`, we can add the synthetic data generation process in CEVAE paper. (Causal Effect Inference with Deep Latent-Variable Models,  http://papers.nips.cc/paper/7223-causal-effect-inference-with-deep-latent-variable-models.pdf)

I noticed that we already have the implementation code under [pyro repo](https://github.com/pyro-ppl/pyro/blob/dev/examples/contrib/cevae/synthetic.py), maybe we can migrate it here to CausalML under [dataset](https://github.com/uber/causalml/tree/master/causalml/dataset)

fritzo:
Resolves #120 

This adds the synthetic dataset from [Luizos et al. 2018](http://papers.nips.cc/paper/7223-causal-effect-inference-with-deep-latent-variable-models.pdf):

image
image

## Tested
- added a cheap unit test to tests/test_datasets.py

### DIFF
--- a/causalml/dataset/__init__.py
+++ b/causalml/dataset/__init__.py
@@ -3,6 +3,7 @@ from .regression import simulate_nuisance_and_easy_treatment
 from .regression import simulate_randomized_trial
 from .regression import simulate_easy_propensity_difficult_baseline
 from .regression import simulate_unrelated_treatment_control
+from .regression import simulate_hidden_confounder
 from .classification import make_uplift_classification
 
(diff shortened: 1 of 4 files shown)


fritzo:
@yungmsh thanks for the help!

Issue #100, opened 2019-12-08
Causalml fails to install xgboost even with xgboost present in my conda env


kgolyaev:
Installing `causalml` through `pip` fails in my environment with the following error:

```
Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/private/var/folders/4f/ck1srdbj21s8j7nbkbdn6zdm0000gn/T/pip-install-n76pirmd/xgboost/setup.py", line 42, in <module>
        LIB_PATH = libpath['find_lib_path']()
      File "/private/var/folders/4f/ck1srdbj21s8j7nbkbdn6zdm0000gn/T/pip-install-n76pirmd/xgboost/xgboost/libpath.py", line 48, in find_lib_path
        'List of candidates:\n' + ('\n'.join(dll_path)))
    XGBoostLibraryNotFound: Cannot find XGBoost Library in the candidate path, did you install compilers and run build.sh in root path?
```

I thought I'd try to install `xgboost` myself, but this did not solve the problem. Here is my system spec:
```
Darwin PCNAME 18.5.0 Darwin Kernel Version 18.5.0: Mon Mar 11 20:40:32 PDT 2019; root:xnu-4903.251.3~3/RELEASE_X86_64 x86_64
```

And here is the conda environment inside which I am trying to install `causalml`:

```
# platform: osx-64
_py-xgboost-mutex=2.0=cpu_0
absl-py=0.8.1=pypi_0
appnope=0.1.0=py37_0
astor=0.8.0=pypi_0
attrs=19.3.0=py_0
backcall=0.1.0=py37_0
blas=1.0=mkl
bleach=3.1.0=py37_0
ca-certificates=2019.11.27=0
certifi=2019.11.28=py37_0
cycler=0.10.0=py37_0
cython=0.29.14=py37h0a44026_0
dbus=1.13.12=h90a0687_0
decorator=4.4.1=py_0
defusedxml=0.6.0=py_0
econml=0.6=pypi_0
entrypoints=0.3=py37_0
expat=2.2.6=h0a44026_0
freetype=2.9.1=hb4e5f40_0
gast=0.2.2=pypi_0
gettext=0.19.8.1=h15daf44_3
glib=2.63.1=hd977a24_0
google-pasta=0.1.8=pypi_0
grpcio=1.25.0=pypi_0
h5py=2.10.0=pypi_0
icu=58.2=h4b95b61_1
importlib_metadata=1.1.0=py37_0
intel-openmp=2019.4=233
[...]

PR #134, opened 2020-02-12
Do not pip install xgboost if already installed

fritzo:
Addresses #100 

This adds a workaround to setup.py to avoid pip installing xgboost if it can already be imported. This works for me when creating a new conda env, e.g.
```sh
conda create -n my_env python=3.6
conda activate my_env
conda install -c conda-forge xgboost
cd causalml
pip install -e .  # works after this PR
```

I am unsure whether this messes up the requirements `pypi` requirements, i.e. when releasing. This may only work when installing locally via `pip install .`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,6 +19,12 @@ with open("README.md", "r") as f:
 
 with open("requirements.txt") as f:
     requirements = f.readlines()
+# workaround to avoid installing xgboost if already installed
+try:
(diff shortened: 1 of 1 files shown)


fritzo:
Closing because I believe this messes up requirements when pip uploading. I'll reopen if we can find a solution. cc @neerajprad

PR #137, opened 2020-02-13
adding filter feature selection methods for uplift modeling

zhenyuz0500:
Adding feature selection filter methods for uplift modeling.
Methods include: F filter, LR filter, KL filter, Chi filter, ED filter (more details to be described by a upcoming paper)

To use these methods:
from causalml.feature_selection import FilterSelect

An example notebook will be added later.

### DIFF
--- a/causalml/feature_selection/filters.py
+++ b/causalml/feature_selection/filters.py
@@ -16,30 +16,32 @@ class FilterSelect:
     def __init__(self):
         return
 
-    def _convert_treatment(self, data, T_name_before, treatment_key):
-        """
-        Convert text treatment labels into 0 (control) and 1 (treated).
-        """
-        T_dummies = pd.get_dummies(data[T_name_before])
-        T_dummies_treatment = T_dummies[treatment_key]   
-        return T_dummies_treatment
-
-
-    def _filter_F_one_feature(self, data, T_name, X_name, y_name): 
+    def _filter_F_one_feature(self, data, treatment_indicator, feature_name, y_name): 
         """
         Conduct F-test of the interaction between treatment and one feature.
+
+        Parameters
+        ----------
+        data (pd.Dataframe): DataFrame containing outcome, features, and experiment group
+        treatment_indicator (string): the column name for binary indicator of treatment (value 1) or control (value 0) 
+        feature_name (string): feature name, as one column in the data DataFrame
+        y_name (string): name of the outcome variable
+
+        Returns
+        ----------
+        (pd.DataFrame): a data frame containing the feature importance statistics
         """
         Y = data[y_name]
-        X = data[[T_name, X_name]]
+        X = data[[treatment_indicator, feature_name]]
         X = sm.add_constant(X)
-        X['{}-{}'.format(T_name, X_name)] = X[[T_name, X_name]].product(axis=1)
+        X['{}-{}'.format(treatment_indicator, feature_name)] = X[[treatment_indicator, feature_name]].product(axis=1)
 
         model = sm.OLS(Y, X)
         result = model.fit()
 
         F_test = result.f_test(np.array([0, 0, 0, 1]))
         F_test_result = pd.DataFrame({
-            'feature': X_name, # for the interaction, not the main effect
+            'feature': feature_name, # for the interaction, not the main effect
             'method': 'F-statistic',
             'score': F_test.fvalue[0][0], 
             'p_value': F_test.pvalue, 
@@ -49,14 +51,25 @@ class FilterSelect:
         return F_test_result
 
 
-    def filter_F(self, data, T_name, X_name_list, y_name):
+    def filter_F(self, data, treatment_indicator, features, y_name):
         """
         Rank features based on the F-statistics of the interaction.
+
+        Parameters
+        ----------
+        data (pd.Dataframe): DataFrame containing outcome, features, and experiment group
+        treatment_indicator (string): the column name for binary indicator of treatment (value 1) or control (value 0) 
+        features (list of string): list of feature names, that are columns in the data DataFrame
+        y_name (string): name of the outcome variable
+
+        Returns
+        ----------
+        (pd.DataFrame): a data frame containing the feature importance statistics
         """
         all_result = pd.DataFrame()
-        for x_name_i in X_name_list: 
+        for x_name_i in features: 
             one_result = self._filter_F_one_feature(data=data,
-                T_name=T_name, X_name=x_name_i, y_name=y_name
+                treatment_indicator=treatment_indicator, feature_name=x_name_i, y_name=y_name
             )
             all_result = pd.concat([all_result, one_result])
 
@@ -66,21 +79,32 @@ class FilterSelect:
         return all_result
 
 
-    def _filter_LR_one_feature(self, data, T_name, X_name, y_name, disp=True): 
+    def _filter_LR_one_feature(self, data, treatment_indicator, feature_name, y_name, disp=True): 
         """
         Conduct LR (Likelihood Ratio) test of the interaction between treatment and one feature.
+
+        Parameters
+        ----------
+        data (pd.Dataframe): DataFrame containing outcome, features, and experiment group
+        treatment_indicator (string): the column name for binary indicator of treatment (value 1) or control (value 0) 
+        feature_name (string): feature name, as one column in the data DataFrame
+        y_name (string): name of the outcome variable
+
+        Returns
+        ----------
+        (pd.DataFrame): a data frame containing the feature importance statistics
         """
         Y = data[y_name]
         
         # Restricted model
-        X_r = data[[T_name, X_name]]
+        X_r = data[[treatment_indicator, feature_name]]
         X_r = sm.add_constant(X_r)
         model_r = sm.Logit(Y, X_r)
         result_r = model_r.fit(disp=disp)
 
         # Full model (with interaction)
         X_f = X_r.copy()
-        X_f['{}-{}'.format(T_name, X_name)] = X_f[[T_name, X_name]].product(axis=1)
+        X_f['{}-{}'.format(treatment_indicator, feature_name)] = X_f[[treatment_indicator, feature_name]].product(axis=1)
         model_f = sm.Logit(Y, X_f)
         result_f = model_f.fit(disp=disp)
 
@@ -89,7 +113,7 @@ class FilterSelect:
         LR_pvalue = 1 - stats.chi2.cdf(LR_stat, df=LR_df)
 
         LR_test_result = pd.DataFrame({
-            'feature': X_name, # for the interaction, not the main effect
+            'feature': feature_name, # for the interaction, not the main effect
             'method': 'LRT-statistic',
             'score': LR_stat, 
             'p_value': LR_pvalue,
@@ -99,14 +123,25 @@ class FilterSelect:
         return LR_test_result
 
 
-    def filter_LR(self, data, T_name, X_name_list, y_name, disp=True):
+    def filter_LR(self, data, treatment_indicator, features, y_name, disp=True):
         """
         Rank features based on the LRT-statistics of the interaction.
+
+        Parameters
+        ----------
+        data (pd.Dataframe): DataFrame containing outcome, features, and experiment group
+        treatment_indicator (string): the column name for binary indicator of treatment (value 1) or control (value 0) 
+        feature_name (string): feature name, as one column in the data DataFrame
+        y_name (string): name of the outcome variable
+
+        Returns
+        ----------
+        (pd.DataFrame): a data frame containing the feature importance statistics
         """
         all_result = pd.DataFrame()
-        for x_name_i in X_name_list: 
+        for x_name_i in features: 
             one_result = self._filter_LR_one_feature(data=data, 
-                T_name=T_name, X_name=x_name_i, y_name=y_name, disp=disp
+                treatment_indicator=treatment_indicator, feature_name=x_name_i, y_name=y_name, disp=disp
             )
             all_result = pd.concat([all_result, one_result])
 
@@ -121,17 +156,12 @@ class FilterSelect:
                         experiment_group_column='treatment_group_key', 
                         y_name='conversion'):
         """
-        To count the conversions and get the probabilities by treatment groups.
+        To count the conversions and get the probabilities by treatment groups. This function comes from the uplift tree algorithm, that is used for tree node split evaluation.
 
         Parameters
         ----------
         data : DataFrame
             The DataFrame that contains all the data (in the current "node").  
-            - [TODO] We may need a index vector to restrict the rows of data 
-              outside the function.
-            - [TODO] Currently the positions of treatment_group_key and 
-              conversion in data are hard coded.  Modify the code for more
-              flexibility.
 
         Returns
         -------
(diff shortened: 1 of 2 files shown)


causalml/feature_selection/filters.py
```diff
@@ -0,0 +1,370 @@
...
+class FilterSelect:
+    """A class for feature importance methods.
+    """
+
+    def __init__(self):
+        return
+
+    def _convert_treatment(self, data, T_name_before, treatment_key):
+        """
+        Convert text treatment labels into 0 (control) and 1 (treated).
```
t-tte: Could we add the args here?

causalml/feature_selection/filters.py
```diff
@@ -0,0 +1,370 @@
...
+        """
+        T_dummies = pd.get_dummies(data[T_name_before])
+        T_dummies_treatment = T_dummies[treatment_key]   
+        return T_dummies_treatment
+
+
+    def _filter_F_one_feature(self, data, T_name, X_name, y_name): 
+        """
+        Conduct F-test of the interaction between treatment and one feature.
```
t-tte: Could we add the args here?

causalml/feature_selection/filters.py
```diff
@@ -0,0 +1,370 @@
+"""
+Filter feature selection methods for uplift modeling
```
t-tte: Perhaps move to main docstring.

causalml/feature_selection/filters.py
```diff
@@ -0,0 +1,370 @@
...
+            'p_value': F_test.pvalue, 
+            'misc': 'df_num: {}, df_denom: {}'.format(F_test.df_num, F_test.df_denom), 
+        }, index=[0]).reset_index(drop=True)
+
+        return F_test_result
+
+
+    def filter_F(self, data, T_name, X_name_list, y_name):
+        """
+        Rank features based on the F-statistics of the interaction.
```
t-tte: Could we add the args here?

causalml/feature_selection/filters.py
```diff
@@ -0,0 +1,370 @@
...
+
+        all_result = all_result.sort_values(by='score', ascending=False)
+        all_result['rank'] = all_result['score'].rank(ascending=False)
+
+        return all_result
+
+
+    def _filter_LR_one_feature(self, data, T_name, X_name, y_name, disp=True): 
+        """
+        Conduct LR (Likelihood Ratio) test of the interaction between treatment and one feature.
```
t-tte: Could we add the args here?

causalml/feature_selection/filters.py
```diff
@@ -0,0 +1,370 @@
...
+            'p_value': LR_pvalue,
+            'misc': 'df: {}'.format(LR_df), 
+        }, index=[0]).reset_index(drop=True)
+
+        return LR_test_result
+
+
+    def filter_LR(self, data, T_name, X_name_list, y_name, disp=True):
+        """
+        Rank features based on the LRT-statistics of the interaction.
```
t-tte: Could we add the args here?

causalml/feature_selection/filters.py
```diff
@@ -0,0 +1,370 @@
...
+            Counts of conversions by treatment groups, of the form: 
+            {'control': {0: 10, 1: 8}, 'treatment1': {0: 5, 1: 15}}
+        nodeSummary: dict
+            Probability of conversion and group size by treatment groups, of 
+            the form:
+            {'control': [0.490, 500], 'treatment1': [0.584, 500]}
+
+        References
+        ----------
+        - results: upliftpy.UpliftTreeClassifier.group_uniqueCounts()
```
t-tte: Are these in the wrong section?

causalml/feature_selection/filters.py
```diff
@@ -0,0 +1,370 @@
...
+        treatment_group_keys = results_series.index.levels[0].tolist()
+        y_name_keys = results_series.index.levels[1].tolist()
+
+        results = {}
+        for ti in treatment_group_keys: 
+            results.update({ti: {}}) 
+            # results[ti] = {} # alternatively
+            for ci in y_name_keys:
+                results[ti].update({ci: results_series[ti, ci]}) 
+                # results[ti][ci] = results_series[ti, ci] # alternatively
```
t-tte: Consider removing.

causalml/feature_selection/filters.py
```diff
@@ -0,0 +1,370 @@
...
+    def _kl_divergence(self, pk, qk):
+        """
+        Calculate KL Divergence for binary classification.
+
+        Parameters
+        ----------
+        pk, qk : float
+        """
+        # Formula
+        # S = sum(np.array(pk) * np.log(np.array(pk) / np.array(qk)))
```
t-tte: Consider removing.

causalml/feature_selection/filters.py
```diff
@@ -0,0 +1,370 @@
...
+        pc = nodeSummary[control_group][0]
+        d_res = 0
+        for treatment_group in nodeSummary:
+            if treatment_group != control_group:
+                d_res += self._kl_divergence(nodeSummary[treatment_group][0], pc)
+        return d_res
+
+    def _evaluate_ED(self, nodeSummary, control_group='control'):
+        """
+        Calculate the multi-treatment unconditional D (one node)
```
t-tte: Could we add the args here?

causalml/feature_selection/filters.py
```diff
@@ -0,0 +1,370 @@
...
+        d_res = 0
+        for treatment_group in nodeSummary:
+            if treatment_group != control_group:
+                d_res += 2 * (nodeSummary[treatment_group][0] - pc)**2
+        return d_res
+
+    def _evaluate_Chi(self, nodeSummary, control_group='control'):
+        """
+        Calculate the multi-treatment unconditional D (one node)
+        with Chi-Square as split Evaluation function.
```
t-tte: Could we add the args here?

causalml/feature_selection/filters.py
```diff
@@ -0,0 +1,370 @@
...
+            'misc': 'number_of_bins: {}'.format(min(n_bins, x_bin.max()+1)),# format(n_bins),
+        }, index=[0]).reset_index(drop=True)
+
+        return(D_result)
+
+    def filter_D(self, data, X_name_list, y_name, 
+                 n_bins=10, method='KL', control_group='control',
+                 experiment_group_column='treatment_group_key'):
+        """
+        Rank features based on the chosen divergence measure.
```
t-tte: Could we add the args here?

causalml/feature_selection/filters.py
```diff
@@ -0,0 +1,370 @@
...
+
+        all_result = all_result.sort_values(by='score', ascending=False)
+        all_result['rank'] = all_result['score'].rank(ascending=False)
+
+        return all_result
+
+
+    # Get node summary - a function 
+    def _GetNodeSummary(self, data, 
+                        experiment_group_column='treatment_group_key',
```
t-tte: Should we consider another default argument than 'treatment_group_key'?

causalml/feature_selection/filters.py
```diff
@@ -0,0 +1,370 @@
...
+        # Counts of conversions by treatment group
+        results_series = data.groupby([experiment_group_column, y_name]).size()
+        
+        treatment_group_keys = results_series.index.levels[0].tolist()
+        y_name_keys = results_series.index.levels[1].tolist()
+
+        results = {}
+        for ti in treatment_group_keys: 
+            results.update({ti: {}}) 
+            # results[ti] = {} # alternatively
```
t-tte: Consider removing.

causalml/feature_selection/filters.py
```diff
@@ -0,0 +1,370 @@
...
+        # Formula
+        # S = sum(np.array(pk) * np.log(np.array(pk) / np.array(qk)))
+        if qk < 0.1**6:
+            qk = 0.1**6
+        elif qk > 1 - 0.1**6:
+            qk = 1 - 0.1**6
+        S = pk * np.log(pk / qk) + (1-pk) * np.log((1-pk) / (1-qk))
+        return S
+
+    def _evaluate_KL(self, nodeSummary, control_group='control'):
```
t-tte: Same question as above regarding the default argument. I think we use 0 as the default control label elsewhere.

causalml/feature_selection/filters.py
```diff
@@ -0,0 +1,370 @@
...
+                d_res += (
+                    (nodeSummary[treatment_group][0] - pc)**2 / max(0.1**6, pc) 
+                    + (nodeSummary[treatment_group][0] - pc)**2 / max(0.1**6, 1-pc)
+                )
+        return d_res
+
+
+    def _filter_D_one_feature(self, data, X_name, y_name, 
+                              n_bins=10, method='KL', control_group='control',
+                              experiment_group_column='treatment_group_key'):
```
t-tte: As above, consider changing the default argument.

causalml/feature_selection/filters.py
```diff
@@ -0,0 +1,370 @@
...
+        if method == 'KL':
+            evaluationFunction = self._evaluate_KL
+        elif method == 'ED':
+            evaluationFunction = self._evaluate_ED
+        elif method == 'Chi':
+            evaluationFunction = self._evaluate_Chi
+
+        totalSize = len(data.index)
+        x_bin = pd.qcut(data[X_name].values, n_bins, labels=False, 
+                        duplicates='drop')
```
t-tte: Do we want to silently drop the duplicates or eg add some jitter? Recall the problem with small sample sizes.

causalml/feature_selection/filters.py
```diff
@@ -0,0 +1,370 @@
...
+            'score': d_res, 
+            'p_value': None,
+            'misc': 'number_of_bins: {}'.format(min(n_bins, x_bin.max()+1)),# format(n_bins),
+        }, index=[0]).reset_index(drop=True)
+
+        return(D_result)
+
+    def filter_D(self, data, X_name_list, y_name, 
+                 n_bins=10, method='KL', control_group='control',
+                 experiment_group_column='treatment_group_key'):
```
t-tte: The default arg :)

causalml/feature_selection/filters.py
```diff
@@ -0,0 +1,370 @@
...
+            )
+            all_result = pd.concat([all_result, one_result])
+
+        all_result = all_result.sort_values(by='score', ascending=False)
+        all_result['rank'] = all_result['score'].rank(ascending=False)
+
+        return all_result
+
+    def get_importance(self, data, X_name_list, y_name, method, 
+                      experiment_group_column='treatment_group_key',
```
t-tte: Same as above regarding the default arg.

causalml/feature_selection/filters.py
```diff
@@ -0,0 +1,370 @@
...
+                experiment_group_column=experiment_group_column, 
+            )
+            all_result = pd.concat([all_result, one_result])
+
+        all_result = all_result.sort_values(by='score', ascending=False)
+        all_result['rank'] = all_result['score'].rank(ascending=False)
+
+        return all_result
+
+    def get_importance(self, data, X_name_list, y_name, method,
```
t-tte: More generally, I think it would make sense to use same conventions as elsewhere in the package, ie use "X", "treatment", "y" and "control_name" as the argument names.

causalml/feature_selection/filters.py
zhenyuz0500: This is an intermediate function that is not user-facing. It seems sklearn skips the args for intermediate functions (with _ prefix), for example: https://github.com/scikit-learn/scikit-learn/blob/b194674c4/sklearn/ensemble/_forest.py#L2172. But we can make our own rules. If we decide to add args for all functions, I will add args here.

causalml/feature_selection/filters.py
zhenyuz0500: adding descriptions to .rst file?

causalml/feature_selection/filters.py
zhenyuz0500: any suggestions?

causalml/feature_selection/filters.py
zhenyuz0500: good catch, the references need to be updated.

causalml/feature_selection/filters.py
zhenyuz0500: agree

causalml/feature_selection/filters.py
zhenyuz0500: agree

causalml/feature_selection/filters.py
zhenyuz0500: The current data generation function generates 'control' and 'treatment', and this setting would compile with the DGP well. However, this is an intermediate function, that is not user-facing.

causalml/feature_selection/filters.py
zhenyuz0500: Good call. One difficulty is that the scale of X is unknown, therefore, adding 10**(-8) may not be always safe. One option is to leave it to users to manipulate X outside, the other option, is to add a value that is proportional to the scale of X, e.g. at the scale of 10**(-8)*median(X).

causalml/feature_selection/filters.py
zhenyuz0500: hmm, one difference is that, in other functions, X and y are values in list format, but X_name_list and y_name is string here. The consideration is that feature importance will come with feature name as outputs for interpretation, therefore only values will be insufficient. Instead, we require inputs in (data, X_name_list, y_name) format.

causalml/feature_selection/filters.py
t-tte: I'm also fine with the convention of not specifying the args for non-public functions. I was purely thinking it from the point of view that it was not immediately clear to me what "T_name_before" means.

causalml/feature_selection/filters.py
t-tte: We're using 0 elsewhere in the package. Not sure that it matters that much if it is a non-public function.

causalml/feature_selection/filters.py
t-tte: NVM, I think it's fine as it is.

causalml/feature_selection/filters.py
t-tte: True. In terms of pre-existing convention, it looks like we're using 'features' for a list of feature names in the explainer class here: https://github.com/uber/causalml/blob/master/causalml/inference/meta/explainer.py For y_name, I'm not sure that any convention exists.

causalml/feature_selection/filters.py
t-tte: In the interest of explicitness, I think it may make sense to switch to duplicates='raise' and let the user manipulate X as needed.

causalml/feature_selection/filters.py
jeongyoonlee: I second to @t-tte. When it might not be straightforward to others, it's better to have docstrings.

causalml/feature_selection/filters.py
zhenyuz0500: hmm, is 0 a default column name, or a default value within the experiment label column? To clarify, experiment_group_column is a column name, and data[experiment_group_column] will be the list of control / treatment labels.

causalml/feature_selection/filters.py
zhenyuz0500: Updated "T_name" to a more interpretable name "treatment_indicator". This arg will only be used for internal functions. Args are added below.

causalml/feature_selection/filters.py
zhenyuz0500: done

causalml/feature_selection/filters.py
zhenyuz0500: done

causalml/feature_selection/filters.py
zhenyuz0500: done

causalml/feature_selection/filters.py
zhenyuz0500: done

causalml/feature_selection/filters.py
zhenyuz0500: done

Issue #156, opened 2020-03-17
Model fails to converge if no propensity scores are provided for X and R Learners


steveyang90:
**Describe the bug**
If propensity score is not provided for X and R Learners, the fit method hangs with the output `The max_iter was reached which means the coef_ did not converge` repeatedly.

**To Reproduce**
```python
learner_x = BaseXClassifier(
    learner=LGBMClassifier(),
    control_effect_learner=LinearRegression(),
    treatment_effect_learner=LinearRegression()
)

train_cate_x = learner_x.fit_predict(X=X_train, treatment=treatment_train, y=y_train, p=None)
```

**Expected behavior**
I expect the model to converge and continue on to the next step

**Screenshots**

**Environment (please complete the following information):**
 - OS: macOS
 - Python Version: 3.6
 - Versions of Major Dependencies (`pandas`, `scikit-learn`, `cython`): [`pandas==0.25`, `scikit-learn==0.22`, `cython==0.29`]

**Additional context**
I am able to workaround if I provide propensity as a fixed value of `0.5` for each training / validation row of data.

Also to note if it makes a difference: My data is based on true experimental data so the treatment/control assignment is true random with a balanced split.

After interrupting the kernel from multiple repeated `The max_iter was reached which means the coef_ did not converge` messages, I see the following trace. It seems apparent from the trace that the convergence issue is occurring during the fitting of the propensity model.

```python
---------------------------------------------------------------------------
KeyboardInterrupt                         Traceback (most recent call last)
<ipython-input-241-9901562a5bd2> in <module>
      2 
      3 # there's a bug if we don't provide propensity
----> 4 train_cate_x = learner_x.fit_predict(X=X_train, treatment=treatment_train, y=y_train, p=None)

/dsw/snapshots/snapshot_dsw_default_jupyter/python3/lib/python3.6/site-packages/causalml/inference/meta/xlearner.py in fit_predict(self, X, treatment, y, p, return_ci, n_bootstraps, bootstrap_size, return_components, verbose)
    240         """
    241         X, treatment, y = convert_pd_to_np(X, treatment, y)
--> 242         self.fit(X, treatment, y, p)
    243 
    244         if p is None:

/dsw/snapshots/snapshot_dsw_default_jupyter/python3/lib/python3.6/site-packages/causalml/inference/meta/xlearner.py in fit(self, X, treatment, y, p)
    625                 w = (treatment == group).astype(int)
[...]

PR #157, opened 2020-03-23
update match.py (scale matching distances) and propensity.py (address issue #156)

yungmsh:
**match.py changes:**
- adjust matching distances by a factor of the number of matching columns (otherwise user has to tune caliper according to how many matching columns they provide)
- use np.array instead of lists for matched idxs

**propensity.py changes**
- add function that checks if data is randomized using AUC (if close to 0.5, bypasses LogisticRegressionCV to prevent repeated non-convergence warnings)
- removed redundant manual cv code in compute_propensity_score (cv is already implemented inside LogisticRegressionCV)

### DIFF
--- a/causalml/propensity.py
+++ b/causalml/propensity.py
@@ -52,9 +52,6 @@ class ElasticNetPropensityModel(object):
             X (numpy.ndarray): a feature matrix
             y (numpy.ndarray): a binary target vector
         """
-        if self.check_if_randomized(X, y):
-            logger.warning('Propensity AUC close to 0.5. This may suggest your data is already randomized.')
-            self.model = LogisticRegression()
         self.model.fit(X, y)
 
     def predict(self, X):
@@ -84,16 +81,6 @@ class ElasticNetPropensityModel(object):
         ps = self.predict(X)
         logger.info('AUC score: {:.6f}'.format(auc(y, ps)))
         return ps
-
-    def check_if_randomized(self, X, y, max_auc=0.51):
-        """Checks if the underlying data is already randomized by fitting a single LogisticRegression.
-        """
-        lr = LogisticRegression()
-        lr.fit(X, y)
-        score = auc(y, lr.predict_proba(X)[:, 1])
-        if score < max_auc:
-            return True
-        return False
 
 def calibrate(ps, treatment):
     """Calibrate propensity scores with logistic GAM.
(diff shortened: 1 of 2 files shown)


causalml/propensity.py
```diff
@@ -47,6 +52,9 @@ def fit(self, X, y):
             X (numpy.ndarray): a feature matrix
             y (numpy.ndarray): a binary target vector
         """
+        if self.check_if_randomized(X, y):
```
steveyang90: @yungmsh what are your thoughts on adding a boolean to the meta learner class itself that can bypass this all-together if user explicitly inputs something like `is_randomized_treatment` or `is_experimental` or something like that? It could possibly bypass a lot of compute time?

yungmsh: Not a bad idea- we could add that so it bypasses it altogether, and even if user doesn't provide an argument for is_randomized_treatment (presumably defaulted to None),under the hood we can do a check anyway to prevent the repeated non-convergence situation

ppstacy:
@yungmsh Just curious, will the convergence error always happen when AUC close to 0.5 or it's just this case for the dataset Steve used? Also wondering this maybe you already tired so increase `max_iter` parameter didn't help to solve this?

causalml/propensity.py
yungmsh: @steveyang90 I thought about it again and think it's better to leave it to the user to provide a constant vector of propensities, and in the case the user doesn't provide it, it'll just do a single fit before detecting that it should be 0.5. Main reasons being i) additional code complexity doesn't justify minimal burden placed on the user - our code would have to account for both single treatment case (vector) and multiple treatment case (dict), and ii) leaving these learners as abstracted as possible is a positive thing. Happy to hear others thoughts here though.

yungmsh:
@ppstacy the convergence error that @steveyang90 was referring to was due to the fact that the underlying data is randomized to begin with. Increasing max_iter may help in cases where there is still more "learning" for the model to do, but this PR addresses Steve's case, in which increasing max_iter theoretically wouldn't solve the issue. If we see cases in the future where users are experiencing non-convergence with non-randomized data, we can recommend them to build a propensity model outside of the base learner, where they may have more control over the model (i.e. be able to tweak max_iter).

yungmsh:
@steveyang90 we just had a meeting about this and Jeong recommended that we hold off on the hacky AUC approach to detecting randomization and address this issue in a later PR.

Issue #190, opened 2020-06-05
IIA Integration: Sensitivity Analysis


jeongyoonlee:
Migrate IIA's sensitivity analysis into CausalML

PR #199, opened 2020-06-17
Sensitivity analysis

ppstacy:
#190 
There are 5 functions in IIA package for sensitivity analysis:

- `data_subset()` is the same as `SensitivitySubsetData()`

- `pseudo_intervention()` is same as `SensitivityPlaceboTreatment()`

- `random_replacement()` is same as `SensitivityRandomReplace()`

- `random_covariate()` is same as `SensitivityRandomCause()`

- `unmeasured_confounding()` is same as `SensitivitySelectionBias()`


causalml/metrics/sensitivity.py
```diff
@@ -0,0 +1,475 @@
...
+import pandas as pd
+from collections import defaultdict
+import matplotlib.pyplot as plt
+from importlib import import_module
+from causalml.inference.meta import BaseTLearner
+
+logger = logging.getLogger('sensitivity')
+
+
+class ConfoundingFunction(object):
```
jeongyoonlee, line 12: We don't need the `ConfoundingFunction` class. Let's make 4 confounding functions available directly and define a catalog dictionary with the `str` name as key and the function as value so that `Sensitivity.sensitivity_analysis()` and `SensitivitySelectionBias.__init__()` take `str` for the `confound` input as follows:
```python
confounding_functions = {'one_sided': one_sided, ...}
...
    def sensitivity_analysis(self, methods, sample_size=None,
                                            confound='one_sided', alpha_range=None):
        ...
        try:
            confound_func = confounding_functions[confound]
        except KeyError:
            raise NotImplementedError(f'Confounding function, {confound} is not implemented. Use one of {confounding_functions.keys()}')
```

causalml/metrics/sensitivity.py
```diff
@@ -0,0 +1,475 @@
+import logging
+import numpy as np
+import pandas as pd
+from collections import defaultdict
+import matplotlib.pyplot as plt
+from importlib import import_module
+from ..inference.meta.tlearner import BaseTLearner
```
jeongyoonlee, line 7: Currently this causes circular import and ImportError in Travis build. We can resolve it by importing it inside the `Sensitivity.get_ate_ci()` right before when it is used.

ppstacy: Thanks @jeongyoonlee! It works!

Issue #211, opened 2020-07-17
example for feature selection


reneix:
Is there any example code for feature selection? I'm using UpliftRandomForestClassifier
I saw this , but still not sure how to do filter works...
Feature Selection Methods for Uplift Modeling , https://arxiv.org/pdf/2005.03447.pdf
thx

paullo0106:
Had a chance to catch up with @yungmsh , will create an example notebook based on their previous work

yungmsh:
Thanks @paullo0106 !

PR #223, opened 2020-08-04
Add feature selection example with Filter methods for issue #211

paullo0106:
This is to address [#211](https://github.com/uber/causalml/issues/211), the notebook can be viewed via [this link](https://github.com/paullo0106/causalml/blob/feature_selection/examples/feature_selection.ipynb) and it includes:

- Demonstrate how to use Filter methods to select the most important features
- Evaluate and compare the AUUC performance with top features dataset


yungmsh:
Looks great, thanks for putting this notebook together! 

2 minor comments:
1. Why don't you use the existing `plot_gain` and `auuc_score` functions instead of defining your own functions in the notebook? Esp if it's an example notebook, I think it would help the reader get acquainted with using the existing functions in the module. Take a look at the last couple cells in https://github.com/uber/causalml/blob/master/examples/dragonnet_example.ipynb
2. Let's combine the uplift curves into a single chart - the above link will explain what I mean. I think we can probably do all_features vs top_10_f vs top_15_f in a single chart, one for each learner.

paullo0106:
Thanks @yungmsh ! 

1. I actually call `get_cumgain` and `plot_gain` in the added functions `plot_uplift` and `get_auuc_score`, I wrapped a few lines in the functions just to avoid repeated code in my notebook. But I see your point, let me see if I can do anything to make it simpler and more intuitive.

2. Good idea, will work on it.

yungmsh:
Looks great, thanks for the updates. Let's merge!

PR #233, opened 2020-09-13
Implement propensity model abstraction for common interface

waltherg:
This PR intends to provide a common interface for propensity models and tidy up some stuff in the corresponding module.

Also, renames former ElasticNet propensity model while providing backwards-compatible synonym.

### DIFF
--- a/causalml/propensity.py
+++ b/causalml/propensity.py
@@ -1,3 +1,4 @@
+from abc import ABCMeta, abstractproperty
 import logging
 import numpy as np
 from pygam import LogisticGAM, s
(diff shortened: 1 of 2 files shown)


causalml/propensity.py
```diff
@@ -1,46 +1,32 @@
...
-            self.cv = StratifiedKFold(n_splits=n_fold, shuffle=True, random_state=random_state)
-        else:
-            self.cv = cv
-        self.model = LogisticRegressionCV(penalty='elasticnet', solver='saga', Cs=Cs, l1_ratios=l1_ratios,
-                                          cv=self.cv, random_state=random_state)
         self.clip_bounds = clip_bounds
+        self.model_kwargs = model_kwargs
+        self.model = self._model
+
+    @abstractproperty
```
jeongyoonlee, line 27: `@abstractproperty` is deprecated since Python 3.3. Please use `@property` and `@abstractmethod` as suggested in the [doc](https://docs.python.org/3/library/abc.html#abc.abstractproperty).

causalml/propensity.py
```diff
@@ -1,46 +1,32 @@
...
-                such that: 0 < lower < upper < 1, to avoid division by zero in BaseRLearner.fit_predict() step.
-            cv (int or cross-validation generator): The default cross-validation generator used is Stratified K-Folds.
-                If an integer is provided, then it is the number of folds used.
-            random_state (numpy.random.RandomState or int): RandomState or an int seed
 
-        Returns:
-            None
+        clip_bounds (tuple): lower and upper bounds for clipping propensity scores. Bounds should be implemented
+                such that: 0 < lower < upper < 1, to avoid division by zero in BaseRLearner.fit_predict() step.
+        model_kwargs: Keyword arguments to be passed to the underlying classification model.
```
jeongyoonlee, line 21: Missing indentation. Please align the indentation.

jeongyoonlee (approved):
Thanks for the contribution @waltherg! LGTM. I added minor comments about indentation and `@abstarctproperty` - Since `causalml` requires Python 3.6+, it'd be better to use the updated syntax.

causalml/propensity.py
waltherg: Learned something new. Thank you.

causalml/propensity.py
waltherg: Unsure what alignment you're looking for.

`such that` aligned with `lower and ...` above?

jeongyoonlee: All three lines (lines 19 ~ 21) should have additional indentation to be one level below than `Args:`.

jeongyoonlee: Actually, it's cosmetic, and I'm fine with as-is. Please feel free to merge (squash & merge) this to `main`. Thanks again!

waltherg: Got you. All good. Happy to fix that. Am also a fan of tidy code.

waltherg:
@jeongyoonlee I don't think I have permissions to squish and merge into master ... could you do so please in case all issues are resolved.

jeongyoonlee:
I see. Sure. Will merge it now. Thanks for your contribution and support!

Issue #227, opened 2020-08-20
Maybe wrong return obj in UpliftRandomForestClassifier.predict()


ZhiyuanLIPlus:
**Describe the bug**
At the comments of  function predict()@UpliftRandomForestClassifier, i've got the following information:

        Returns
        -------
        df_res : DataFrame, shape = [num_samples, (num_treatments + 1)]
            A DataFrame containing the predicted delta in each treatment group,
            the best treatment group and the maximum delta.

But at the end of function predict()@UpliftRandomForestClassifier, the code is:

`df_res['max_delta'] = df_res[delta_cols].max(axis=1)`
`return y_pred_list`

I'm just wondering maybe the result should be df_res instead of the y_pred_list? 
Because for me, we've got all promised output () in df_res and nothing but only a simple uplift score in y_pred_list (even without the treatment name in multiple treats usecase).

Thanks a lot~ 
And great jobs by the way, very useful toolkit!

ppstacy:
Hi @ZhiyuanLIPlus, thanks for using `CausalML` and your inputs. Here `y_pred_list` meant to only return the prediction results, if you want to put it in a nicer format you can follow the code below which is also available in the [uplift_trees_with_synthetic_data](https://github.com/uber/causalml/blob/master/examples/uplift_trees_with_synthetic_data.ipynb) example notebook. Now after #236  you can also get `df_res` by passing `full_ouput=True` to the function. Please try it out and let us know if you have any more questions. Thanks!

```
y_pred = uplift_model.predict(df_test[x_names].values)

# Put the predictions to a DataFrame for a neater presentation
result = pd.DataFrame(y_pred, columns=uplift_model.classes_)
```

saiwing-yeung:
I noticed something similar but I am not sure if I should open a new issue. In `predict()@UpliftTreeClassifier` the docstring says

```
        Returns
        -------
        df_res : DataFrame, shape = [num_samples, (num_treatments + 1)]
            A DataFrame containing the predicted delta in each treatment group,
            the best treatment group and the maximum delta.
```

But it returns 

```
        if full_output:
            return treatment_optimal, p_hat_optimal, upliftScores, pred_nodes
        else:
            return treatment_optimal, p_hat_optimal
```

ppstacy:
Thanks @saiwing-yeung for reporting this! I think we can just reopen this issue. Do you want to submit a pull request to fix the docstring?

jeongyoonlee:
Fixed docstrings in #422.

PR #236, opened 2020-10-06
Fix the uplift random forest predict output

ppstacy:
To fix issue #227, return `df_res` when `full_output` is True for more detailed results.


jeongyoonlee (approved):
LGTM.

Issue #229, opened 2020-09-06
Is there code error?


ZSCDumin:
<img width="865" alt="截屏2020-09-06 下午7 28 28" src="https://user-images.githubusercontent.com/19670365/92324818-b4019b00-f077-11ea-8364-7836e6835d65.png">

paullo0106:
This seems to be the implementation from the beginning, @zhenyuz0500 @jeongyoonlee thoughts before I dig deeper?
https://github.com/uber/causalml/blob/master/causalml/inference/tree/models.py#L969

jeongyoonlee:
Thanks @ZSCDumin for catching this. It looks like a bug to me, but I'd like to defer to @zhenyuz0500 and @t-tte. @zhenyuz0500 and @t-tte, could you confirm? Thanks!

paullo0106:
action item: Paul to test the updated algorithm

ZSCDumin:
> Thanks @ZSCDumin for catching this. It looks like a bug to me, but I'd like to defer to @zhenyuz0500 and @t-tte. @zhenyuz0500 and @t-tte, could you confirm? Thanks!

yes, I confirm that is a bug.

paullo0106:
[Updated the line](https://github.com/uber/causalml/compare/master...paullo0106:uplift_tree_update?expand=1) in the `UpliftTreeClassifier` as ZSCDumin pointed out, and ran through [uplift_tree_visualization.ipynb](https://github.com/paullo0106/causalml/blob/uplift_tree_update/examples/uplift_tree_visualization.ipynb) and [uplift_trees_with_synthetic_data.ipynb](https://github.com/paullo0106/causalml/blob/uplift_tree_update/examples/uplift_trees_with_synthetic_data.ipynb) examples again as a basic test (there might be other more precise ways to validate, not sure if it's needed at the moment) 

cc: @jeongyoonlee @zhenyuz0500 @t-tte

jeongyoonlee:
That's great! @paullo0106, could you please submit a PR for your fix? Thanks.

paullo0106:
Fixed in the latest merge, thanks to ZSCDumin again!

PR #238, opened 2020-10-07
Pass in currentNodeSummary rather than parentNodeSummary when growDec…

paullo0106:
…isionTreeFrom() for the right branch as ZSCDumin pointed out, and test with two uplift tree example notebooks
this is for issue https://github.com/uber/causalml/issues/229


jeongyoonlee (approved):
LGTM.

PR #248, opened 2020-11-14
Making Tensorflow an optional extra to save unnecessary download

maccam912:
I am operating in a very low disk space environment, and Tensorflow as a dependency is giving me an extra 500 MB of download I don't need. I have moved it here to an optional extra, so a 'pip install causalml[tensorflow]' will still get it, but an installation of causalml without specifying tensorflow will only require xgboost.

All existing tests run when following CONTRIBUTING.md guidelines.

@ppstacy @paullo0106 @waltherg @jeongyoonlee @ccrndn and please tag anyone else I missed.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.5, 3.6, 3.7, 3.8]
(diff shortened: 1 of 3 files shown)


maccam912:
@zhenyuz0500 Sorry I forgot to tag you in the first list. There might still be others I missed too.

maccam912:
Re-opening with only the necessary commits

PR #250, opened 2020-11-16
Make tensorflow optional

maccam912:
I am operating in a very low disk space environment, and Tensorflow as a dependency is giving me an extra 500 MB of download I don't need. I have moved it here to an optional extra, so a 'pip install causalml[tensorflow]' will still get it, but an installation of causalml without specifying tensorflow will only require xgboost.

All existing tests run when following CONTRIBUTING.md guidelines.

@ppstacy @paullo0106 @waltherg @jeongyoonlee @ccrndn @zhenyuz0500 and please tag anyone else I missed.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,4 +17,3 @@ lightgbm
 pygam
 packaging
 keras
-tensorflow>=1.15.2
--- a/setup.py
+++ b/setup.py
@@ -40,6 +40,9 @@ setup(
     long_description_content_type="text/markdown",
     url="https://github.com/uber/causalml",
     packages=packages,
+    extras_require={
+        "tensorflow": ["tensorflow>=1.15.2"],
+    },
     python_requires=">=3.6",
     classifiers=[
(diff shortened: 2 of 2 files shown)


maccam912:
Related to issue https://github.com/uber/causalml/issues/253

jeongyoonlee:
Thanks for the PR, @maccam912! Although, it makes sense to exclude `tensorflow` in your use case, I am not sure about removing it from the default `pip` installation as this PR will do. 

Instead, what about using `--no-deps` in your use case?

It'd be ideal if there's a way to exclude a certain dependency optionally ,but I couldn't find it.

Let me know what you think.

Thanks!

maccam912:
You spend much more time with causalml than I do, so I defer to your judgement. I can close my PR, and try the --no-deps or use my fork if I need to use the reduced size download.

jeongyoonlee:
Thanks. I'll look forward to more contribution from you! :)

Issue #249, opened 2020-11-16
BaseXClassifier, outcome_learner and effect_learner switched?


Garve:
Hi!

Running causalml 0.9.0 on Windows.

Consider the famous kidney stones example. First, I'm going to create the dataset.

```
data = ((*a, b) for (a, b) in zip(itertools.product([0,1], [0,1], [0,1]), [36, 234, 25, 55, 6, 81, 71, 192]))
df = pd.DataFrame(data, columns=['Open_Surgery?', 'Large_Stone?', 'Favorable?', 'N'])
df = df.loc[df.index.repeat(df['N'])].reset_index(drop=True).drop(columns=['N'])
```

Using Microsoft's dowhy and manual approaches (a manual S-learner), one expects that the ATE is around 0.05. I tried to recompute this using causalml's BaseXClassifier.

The help says:

```
Init signature:
BaseXClassifier(
    outcome_learner=None,
    effect_learner=None,
    control_outcome_learner=None,
    treatment_outcome_learner=None,
    control_effect_learner=None,
    treatment_effect_learner=None,
    ate_alpha=0.05,
    control_name=0,
)
Docstring:      A parent class for X-learner classifier classes.
Init docstring:
Initialize an X-learner classifier.

Args:
    outcome_learner (optional): a model to estimate outcomes in both the control and treatment groups.
        Should be a regressor.
    effect_learner (optional): a model to estimate treatment effects in both the control and treatment groups.
        Should be a classifier.
```

However, if I instantiate it via 

```
x = BaseXClassifier(
    outcome_learner=RandomForestRegressor(),
    effect_learner=RandomForestClassifier()
)
```

as suggested by the help, it doesn't work, because 'RandomForestRegressor() doesn't have a predict_proba method' (obviously).

If I switch the learners to

```
x = BaseXClassifier(
    outcome_learner=RandomForestClassifier(),
    effect_learner=RandomForestRegressor()
)
```

everything works and I get a value around 0.05 as the ATE, which is what I would expect. 

```
x.estimate_ate(
    X=df[['Large_Stone?']],
    treatment=df['Open_Surgery?'],
    y=df['Favorable?']
)
``` 
(Works perfectly!)

In the best case, it's only a typo in the help section.

Anyway, thanks for this great package!

Best
Robert

paullo0106:
Thanks for reaching out and the detailed example! You are correct that this is documentation error which _soodimilanlouei_ also mentioned recently in #244

[...]

PR #251, opened 2020-11-16
1. Correct the documentation that outcome_learner should be classifie…

paullo0106:
…r and effect_learner should be regressor instead for BaseXClassifier

2. Remove unused propensity score numpy conversion

This is for #249

### DIFF
--- a/causalml/inference/meta/xlearner.py
+++ b/causalml/inference/meta/xlearner.py
@@ -115,12 +115,6 @@ class BaseXLearner(object):
         else:
             check_p_conditions(p, self.t_groups)
 
-        if isinstance(p, (np.ndarray, pd.Series)):
-            treatment_name = self.t_groups[0]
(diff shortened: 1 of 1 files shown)


causalml/inference/meta/xlearner.py
```diff
@@ -641,12 +641,6 @@ def fit(self, X, treatment, y, p=None):
         else:
             check_p_conditions(p, self.t_groups)
```
ppstacy, line 643: Thanks @paullo0106 for editing! Quick question, any reason we are dropping line 644 - 649?

paullo0106: yeah so from the code `p` is not being used in `fit()`, it's used in `predict()` and `estimate_ate()` when estimating average treatment effect - this is how the X-learner algorithm works, right?

ppstacy (approved):
Thanks, LGTM!

PR #256, opened 2020-11-18
Refactoring the Code by using String Format Arguments, removing Unnecessary Comprehensions and using `is` to compare Object Types

HarshCasper:
## Description 

This Pull Request does the following: 

- Removing unnecessary comprehensions to improve readability
- Using `is` instead of `==` to compare the type of objects 
- Passing String Format Arguments as Logging Method Parameters

The Issue seems to be very minor but it improves the readability of the Code for the better.

I'm a First-Time Contributor to Uber Open-Source so it would be helpful for me understanding how PR workflow works and help me step up in future contributions as well.


jeongyoonlee (approved):
LGTM. Thanks for the contribution, @HarshCasper!

Issue #270, opened 2020-12-08
Feature selection pd.qcut n_bins duplicate issue


manojbalaji1:
I just wanted to check what was the reason behind settings duplicate='raise' in filter using KL? One of the variable which was given as input had raised error due to duplicates in pd.qcuts, it is not a categorical variable

manojbalaji1:
I can change it to drop duplicate edges as instead of raise, drop fixes this or rank the values then do qcut @paullo0106 what do you say?

manojbalaji1:
@paullo0106 I tested with duplicates='drop', which seems to work for me and seems to solve the problem. If you are ok I will do the changes.

paullo0106:
Thanks @manojbalaji1 , I haven't had a chance to take a closer look. Can you submit a PR and we can potentially have more input from other contributors?  Thank you

PR #271, opened 2020-12-10
Fix FilterSelect:_filter_D_one_feature pd.qcut duplicate edges 

manojbalaji1:
## Proposed changes

When binning values, if the quantile values are the same i.e. duplicates exit in quantile values ex: quantiles =  [0, 0, 1, 1], ideally we are supposed to drop the duplicate edges i.e quantiles = [0, 1] or change the n_bins by reducing it. Since we are taking one single value for n_bins and this is being used for every feature in the data frame passed as input, better to drop duplicates else will throw an error. 

Why would duplicates exist?
There might be some feature that has 0 value for 75% of values but the last quantile may have a non-zero value. 

## Types of changes

## Checklist

## Further comments

Binning data using qcut raises an exception when it results in duplicate values for the bins. In this case, we are either supposed to drop the duplicates or change the values of n_bins to adjust the number of bins to avoid duplicity. This PR solve the issue reported in #270

### DIFF
--- a/causalml/feature_selection/filters.py
+++ b/causalml/feature_selection/filters.py
@@ -324,6 +324,8 @@ class FilterSelect:
             evaluationFunction = self._evaluate_Chi
 
         totalSize = len(data.index)
+
+        # drop duplicate edges in pq.cut result to avoid issues
         x_bin = pd.qcut(data[feature_name].values, n_bins, labels=False, 
                         duplicates='drop')
         d_children = 0
--- a/causalml/feature_selection/filters.py
+++ b/causalml/feature_selection/filters.py
@@ -325,7 +325,7 @@ class FilterSelect:
 
         totalSize = len(data.index)
         x_bin = pd.qcut(data[feature_name].values, n_bins, labels=False, 
-                        duplicates='raise')
+                        duplicates='drop')
         d_children = 0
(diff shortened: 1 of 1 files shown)


ppstacy:
Hi @manojbalaji1! Thank you so much for your contribution.  

Instead of directly dropping the duplicated bin our original code used `duplicates='raise'` which will raise ValueError. This way users can change the input parameter `n_bins` to avoid the issue. This is how I understand the logic behind it so it's not necessary to change to `duplicates='drop'`. Please chime in if this is not correct @zhenyuz0500 @yungmsh. Thanks!

manojbalaji1:
@ppstacy yes, idealistic way is to change the bind, but this n_bins is applied on entire dataframe, if there's only one or two features causing this issue then better to drop. Let's say there's a feature with 75% 0's, 22% 1's rest some other value along with other continuous variables. In this case to solve the duplicates, if we change n_bins value, it affects binning of other continuous variables as well. So I thought it's better to drop duplicates. What do you all think?

ppstacy:
@manofporkbelly Yeah, those make sense. I didn't take into consideration before that we are actually changing `n_bins` on the entire dataset.  So I agree that It's better to drop duplicates this way. Thanks! LGTM!

PR #274, opened 2020-12-15
Fix unreproducible UpliftRandomForestClassifier

RuslanTs:
## Proposed changes
issue #275

UpliftRandomForestClassifier doesn't give the same results when runs with the same data and the same params (seed fixed). Now it's reproducible (wasn't fix random seed in bootstrap).

## Types of changes

## Checklist

- [ ] Lint and unit tests pass locally with my changes

## Further comments

### DIFF
--- a/causalml/inference/tree/models.py
+++ b/causalml/inference/tree/models.py
@@ -1288,8 +1288,7 @@ class UpliftRandomForestClassifier:
         self.feature_importances_ = np.mean(all_importances, axis=0)
         self.feature_importances_ /= self.feature_importances_.sum()  # normalize to add to 1
 
-    @staticmethod
-    def bootstrap(X, treatment, y, tree):
+    def bootstrap(self, X, treatment, y, tree):
         np.random.seed(self.random_state)
         bt_index = np.random.choice(len(X), len(X))
         x_train_bt = X[bt_index]
--- a/causalml/inference/tree/models.py
+++ b/causalml/inference/tree/models.py
@@ -136,10 +136,12 @@ class UpliftTreeClassifier:
         The normalization factor defined in Rzepakowski et al. 2012, correcting for tests with large number of splits
         and imbalanced treatment and control splits
 
+    random_state: int, optional (default=2019)
+        The seed used by the random number generator.
     """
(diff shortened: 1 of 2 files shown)


paullo0106:
Looks good, thanks for the contribution!

Just to confirm, do we also need to set `random.seed` in [growDecisionTreeFrom()](https://github.com/uber/causalml/blob/master/causalml/inference/tree/models.py#L941) when we randomly select a few columns? Or this should be handle by the existing random.seed invoked earlier (?) in `fit()` and `bootstrap()`

ppstacy (approved):

RuslanTs:
> Looks good, thanks for the contribution!
> 
> Just to confirm, do we also need to set `random.seed` in [growDecisionTreeFrom()](https://github.com/uber/causalml/blob/master/causalml/inference/tree/models.py#L941) when we randomly select a few columns? Or this should be handle by the existing random.seed invoked earlier (?) in `fit()` and `bootstrap()`

Yes, you right! I fixed this in the new commit.

paullo0106 (approved):
Thanks!

paullo0106:
`Travis CI - Pull Request Failing after 38m — Build Failed` ->  maybe we can rerun the checks and merge

RuslanTs:
Tbh, test cases are quite strange, because in sum by all users a model could be worse than random.
The goal of uplift-modeling is to find some portion of users to treatment, where treatment should be better than it's absence.
I modified the test a bit to pass, but I'm not sure what test cases should be here

RuslanTs:
Seems I cracked bootstrap and my approach is bad. It has to be the completely different realization of bootstrap (like in sklearn, as an example)

paullo0106:
> Seems I cracked bootstrap and my approach is bad. It has to be the completely different realization of bootstrap (like in sklearn, as an example)

@RuslanTs can you clarify this? are there more changes you'd suggest and add to the PR? Thanks

RuslanTs:
It should be a different approach to deal with bootstrap - to make UpliftRandomForestClassifier reproducible. 
My fixes do it reproducible, but bootstrap doesn't work in this case.

PR #297, opened 2021-02-02
Add propensity_learner to R-learner.

jeongyoonlee:
## Proposed changes

Since propensity modeling takes an important role in R-learner, it'd be better if users can define which learner to use for propensity modeling. This PR includes:

- add `propensity_learner` as an input argument to `BaseRLearner` with `ElasticNetPropensityModel` as default.
- refactor R/X-learners by defining `_set_propensity_models()` and `_format_p()` in `BaseLearner`.

## Types of changes

## Checklist

## Further comments

N/A

### DIFF
--- a/causalml/inference/meta/base.py
+++ b/causalml/inference/meta/base.py
@@ -1,7 +1,14 @@
 from abc import ABCMeta, abstractclassmethod
+import logging
 import numpy as np
+import pandas as pd
 
 from causalml.inference.meta.explainer import Explainer
+from causalml.inference.meta.utils import check_p_conditions, convert_pd_to_np
+from causalml.propensity import compute_propensity_score
+
+
+logger = logging.getLogger('causalml')
 
 
 class BaseLearner(metaclass=ABCMeta):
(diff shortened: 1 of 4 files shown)


causalml/inference/meta/base.py
```diff
@@ -38,6 +45,57 @@ def bootstrap(self, X, treatment, y, p=None, size=10000):
...
+        p = dict()
+        p_model = dict()
+        for group in self.t_groups:
+            mask = (treatment == group) | (treatment == self.control_name)
+            treatment_filt = treatment[mask]
+            X_filt = X[mask]
+            w_filt = (treatment_filt == group).astype(int)
+            w = (treatment == group).astype(int)
+            propensity_model = self.model_p if hasattr(self, 'model_p') else None
+            p[group], p_model[group] = compute_propensity_score(X=X_filt, treatment=w_filt,
```
ppstacy, line 93: Hi @jeongyoonlee a quick question that might not directly be related to this PR. Why do we need two `treatment` and `treatment_pred` here? Aren't those the same thing based on here: https://github.com/uber/causalml/blob/master/causalml/propensity.py#L189?

ppstacy (approved):
LGTM, thanks!

causalml/inference/meta/base.py
jeongyoonlee: In this call to `compute_propensity_score()`, `treatment` is only for the filtered samples with `treatment` belonging to either the control group or treatment group of interest while `treatment_pred` is for the entire samples. These two will be the same for a single treatment, but different for the multiple treatments.

ppstacy: Gotcha I see. Sorry a follow-up question here, why do we need to calibrate it with the entire dataset? I am asking because we are only using treatment group X + control when running the model in https://github.com/uber/causalml/blob/master/causalml/inference/meta/rlearner.py#L111-L118.

Issue #300, opened 2021-02-15
Adding another tree-based approach: The delta-delta-p criterium


jroessler:
**Is your feature request related to a problem? Please describe.**
All of your tree-based uplift approaches are based on either the work by Rzepakowski and Jaroszewicz (2012) or the work by Zhao et al. (2017). Yet, there is another tree-based approach which was introduced by Hansotia & Rukstales in 2002. The authors used a so called delta-delta-p criterium for splitting and training the tree. In short: At each split, they calculate the difference in response rate between treatment and control group in each leaf. Subsequently, they calculate the difference between the differences (that's why they call it delta-delta-p) in the leafs. The split (in particular the variable and its corresponding value) with the largest delta-delta-p value is then selected. I think that the method is very intuitiv and easy to understand and it cannot only serve well as a baseline but it can also outperform other approaches. Thus, it should definitely be a part of the uplift-modeling "toolbox"

**Describe the solution you'd like**
I looked into your code and it seems very easy to integrate the method by Hansotia & Rukstales thanks to your well organized code. I have already implemented the code and would be happy to share it with you via a pull request.

**Additional context**
Hansotia & Rukstales: Incremental value modeling (2002)

t-tte:
This sounds good! Feel free to submit the PR.

jroessler:
@t-tte: Sorry for my late response!

I forgot to mention that the DDP approach works only for binary cases in which customers receive either a treatment or not. I guess it is still relevant but I'll mention it in the documents and I'll probably add a warning if the user wants to apply the DDP approach on a dataset with multiple treatments.

PR #327, opened 2021-05-03
Add delta-delta-p (ddp) tree inference approach

jroessler:
## Proposed changes

As suggested in #300  I implemented the delta-delta-p (in short: DDP) tree based inference method originally proposed in Hansotia and Rukstales: Incremental value modeling (2002). I simply added another evaluationFunction in _causalml/inference/tree/models.py_ called 'DDP'.

## Types of changes

- [ ]  Bugfix (non-breaking change which fixes an issue)
- [x]  New feature (non-breaking change which adds functionality)
- [ ]  Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ]  Documentation Update (if none of the other choices apply)

## Checklist

- [x]  I have read the [CONTRIBUTING](https://github.com/uber/causalml/blob/master/CONTRIBUTING.md) doc
- [x]  I have signed the [CLA](https://cla-assistant.io/uber/causalml)
- [x]  Lint and unit tests pass locally with my changes
- [ ]  I have added tests that prove my fix is effective or that my feature works
- [x]  I have added necessary documentation (if appropriate)
- [x]  Any dependent changes have been merged and published in downstream modules

## Further comments

There was no need to add an additional test for proving that the new tree based inference method works. You can simply use the _test_uplift_trees.py_ test and provide _evaluationFunction='DDP'_ when creating the _UpliftRandomForestClassifier_ and _UpliftTreeClassifier_ instances. I noticed that my test does not pass if _N_SAMPLE = 1000_ because the performance of random targeting was better than the DDP approach. However increasing _IN_SAMPLE_ to, for example, 10.000 worked.

If you need additional information, don't hesitate to ask me!

### DIFF
--- a/causalml/inference/tree/models.py
+++ b/causalml/inference/tree/models.py
@@ -1022,7 +1022,7 @@ class UpliftTreeClassifier:
                     if self.control_name in leftNodeSummary and self.control_name in rightNodeSummary:
                         leftScore1 = evaluationFunction(leftNodeSummary, control_name=self.control_name)
                         rightScore2 = evaluationFunction(rightNodeSummary, control_name=self.control_name)
-                        gain = leftScore1 - rightScore2
+                        gain = np.abs(leftScore1 - rightScore2)
                         gain_for_imp = (len(X_l) * leftScore1 - len(X_r) * rightScore2)
                     else:
                         gain = 0
--- a/causalml/inference/tree/models.py
+++ b/causalml/inference/tree/models.py
@@ -116,7 +116,7 @@ class UpliftTreeClassifier:
     ----------
 
     evaluationFunction : string
-        Choose from one of the models: 'KL', 'ED', 'Chi', 'CTS'.
+        Choose from one of the models: 'KL', 'ED', 'Chi', 'CTS', 'DDP'.
 
     max_features: int, optional (default=None)
         The number of features to consider when looking for the best split.
@@ -156,6 +156,8 @@ class UpliftTreeClassifier:
             self.evaluationFunction = self.evaluate_ED
         elif evaluationFunction == 'Chi':
             self.evaluationFunction = self.evaluate_Chi
+        elif evaluationFunction == 'DDP':
+            self.evaluationFunction = self.evaluate_DDP
         else:
             self.evaluationFunction = self.evaluate_CTS
         self.fitted_uplift_tree = None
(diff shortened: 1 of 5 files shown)


causalml/inference/tree/models.py
```diff
@@ -990,6 +1018,14 @@ def growDecisionTreeFrom(self, X, treatment, y, evaluationFunction, max_depth=10
                     rightScore2 = evaluationFunction(rightNodeSummary)
                     gain = (currentScore - p * leftScore1 - (1 - p) * rightScore2)
                     gain_for_imp = (len(X) * currentScore - len(X_l) * leftScore1 - len(X_r) * rightScore2)
+                elif evaluationFunction == self.evaluate_DDP:
+                    if self.control_name in leftNodeSummary and self.control_name in rightNodeSummary:
+                        leftScore1 = evaluationFunction(leftNodeSummary, control_name=self.control_name)
+                        rightScore2 = evaluationFunction(rightNodeSummary, control_name=self.control_name)
+                        gain = np.abs(leftScore1 - rightScore2)
+                        gain_for_imp = (len(X_l) * leftScore1 - len(X_r) * rightScore2)
```
paullo0106, line 1026: @jroessler I'm not familiar with this method, but just want to confirm if it's the case that `gain` takes the absolute difference while `gain_for_imp` doesn't

jroessler: @paullo0106 Thanks for the pointer! Actually it's a mistake! It should be the absolute value for `gain_for_imp` as well. I'll fix it and resubmit.

jroessler:
I added raising an exception if the DDP approach is used with more than two treatment options as mentioned here:
https://github.com/uber/causalml/issues/300#issuecomment-831209598

Further, regarding the tests, I added a method called `generate_classification_data_two_treatments` which creates a synthetic data set with only two treatment options. You can use this method within `test_UpliftTreeClassifier` and `test_UpliftRandomForestClassifier` to verify that the DDP approach is working as excepted. _As a side note_: I realized that the results for all evaluation functions (not only for the DDP approach but also for KL, ED, and CHI) changed from test (RandomForestTest) to test. Thus, it seems that something with `random_state` is not working properly.

jeongyoonlee:
Thanks, @jroessler for the contribution.  It looks good to me. @t-tte, could you check this PR?

t-tte (approved):
Thanks for the contribution.

LGTM also.

Issue #348, opened 2021-06-08
Filter methods fail with NaNs in the data


bponedel:
The filter methods do not work on data containing NaNs, but the error that is generated is not obviously related to this problem. Would be nice to either check for NaNs and error with a good reason or have an option to impute something, maybe a median? Since XGBoost models handle NaNs we typically leave them as is and ran into this problem.

**To Reproduce**
Here is a minimal example:
```
from causalml.feature_selection.filters import FilterSelect
import pandas as pd
import numpy as np

df = pd.DataFrame({
    'feature': np.concatenate([np.arange(80),np.nan*np.ones(20)]),
    'label': np.concatenate([np.ones(60), np.zeros(40)]),
    'treatment': np.concatenate([np.zeros(40), np.ones(60)]).astype(int).astype(str),
})

filter_f = FilterSelect()
df_rank = filter_f.get_importance(
    data=df,
    features=['feature'],
    y_name="label",
    method="KL",
    experiment_group_column='treatment',
    control_group="0",
    treatment_group="1",
)
```
The full traceback is:
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-20-fa53aaf5fcfe> in <module>
     17     experiment_group_column='treatment',
     18     control_group="0",
---> 19     treatment_group="1",
     20 )

~/.pyenv/versions/flex-extrapay/lib/python3.7/site-packages/causalml/feature_selection/filters.py in get_importance(self, data, features, y_name, method, experiment_group_column, control_group, treatment_group, n_bins)
    458                 features=features, y_name=y_name,
    459                 n_bins=n_bins, control_group=control_group,
--> 460                 experiment_group_column=experiment_group_column,
    461             )
    462 

~/.pyenv/versions/flex-extrapay/lib/python3.7/site-packages/causalml/feature_selection/filters.py in filter_D(self, data, features, y_name, n_bins, method, control_group, experiment_group_column)
    402                 data=data, feature_name=x_name_i, y_name=y_name,
    403                 n_bins=n_bins, method=method, control_group=control_group,
--> 404                 experiment_group_column=experiment_group_column,
    405             )
    406             all_result = pd.concat([all_result, one_result])

~/.pyenv/versions/flex-extrapay/lib/python3.7/site-packages/causalml/feature_selection/filters.py in _filter_D_one_feature(self, data, feature_name, y_name, n_bins, method, control_group, experiment_group_column)
    342                         duplicates='drop')
[...]

PR #352, opened 2021-06-10
fix issue filter method failure with NaNs in the data issue #348, add…

manojbalaji1:
…  null_impute using sklearn SimpleImputer

## Proposed changes

NaNs present in the data caused pd.qcut at https://github.com/uber/causalml/blob/master/causalml/feature_selection/filters.py#L341 to return np.nan as NaN values don't fall in any specified bin range. This caused issue since x_bins.max() returned np.nan which caused error at line https://github.com/uber/causalml/blob/master/causalml/feature_selection/filters.py#L344

## Types of changes

## Checklist

## Further comments

First to fix the issue raised in #348 ,  changed the code at https://github.com/uber/causalml/blob/master/causalml/feature_selection/filters.py#L344 from x_bins.max() to np.nanmax(x_bins).astype(int) which will no more return np.nan as max value since it picks max value from given input ignoring np.nan.
Also introduced NaN imputation. As suggested by the author in the aforementioned issue #348, Added a way to pass null_impute argument to impute data. Used SimpleImputer from sklearn to do the same

### DIFF
--- a/causalml/feature_selection/filters.py
+++ b/causalml/feature_selection/filters.py
@@ -8,6 +8,7 @@ import numpy as np
 import pandas as pd
 import statsmodels.api as sm
 from scipy import stats
+from sklearn.impute import SimpleImputer
     
 class FilterSelect:
     """A class for feature importance methods.
@@ -304,7 +305,8 @@ class FilterSelect:
 
     def _filter_D_one_feature(self, data, feature_name, y_name, 
                               n_bins=10, method='KL', control_group='control',
-                              experiment_group_column='treatment_group_key'):
+                              experiment_group_column='treatment_group_key',
+                              null_impute=None):
         """
         Calculate the chosen divergence measure for one feature.
 
@@ -322,6 +324,8 @@ class FilterSelect:
         experiment_group_column (string, optional, default = 'treatment_group_key'): the experiment column name in the DataFrame, which contains the treatment and control assignment label
         control_group (string, optional, default = 'control'): name for control group, value in the experiment group column
         n_bins (int, optional, default = 10): number of bins to be used for bin-based uplift filter methods
+        null_impute (str, optional, default=None): impute np.nan present in the data taking on of the following strategy values {'mean', 'median', 'most_frequent', None}
+        
         Returns
         ----------
         (pd.DataFrame): a data frame containing the feature importance statistics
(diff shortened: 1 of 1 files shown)


paullo0106 (approved):
LGTM, thanks!

manojbalaji1:
Thanks @paullo0106 !

bponedel:
The imputation feature is great but is the resolution of the nan error in the `.max()` really the desired behavior? This will ignore rows of the data that have nans in that column silently and will not raise any error. Would it not be more desirable to raise an error or at least a warning? Just a thought.

Could also link the issue to the PR in the "linked issues" section so that they are associated. 👍

manojbalaji1:
@bponedel 
Shouldn't it be so? Let's say in real life, as a data scientist, If I give you data with nulls, you can either ignore the nulls and go ahead or impute nulls and go ahead. I agree that we should add some warning stating "nulls present in data, please use null_impute to impute null else will be ignored".
@paullo0106 what's your opinion on this?

bponedel:
I see your point and I certainly defer to people more knowledgable about these filter methods and a warning is good.
I guess I was thinking that if the method really doesn't work as intended when rows are skipped (i.e. if it has the chance to misrank the feature) then it might be better to error in the presence of NaNs unless an option is passed to explicitly go forward skipping NaNs. So something like: error if NaNs are present unless an imute arg or `null_impute ="skip"` is passed. The thing that worries me is that the warnings could be too quiet and someone who is new to the method misses it.

paullo0106:
@t-tte adding you to the reviewer list as you were one of the authors of Feature Selection methods with @zhenyuz0500 and @yungmsh - your input would be very helpful here!

@manojbalaji1 on a side note, if we are pushing this through, we might also need to replace line 367 `'misc': 'number_of_bins: {}'.format(min(n_bins, x_bin.max()+1))` with your `x_bin.nanmax()` changes

t-tte:
Thanks for the heads up @paullo0106. If there are NaNs present in the features, I think it's safest to raise the following error:

> NaNs present in features. Please specify null_impute argument or deal with NaNs before using the method.

We don't currently have a good idea of how skipping rows (or, for that matter, different imputation methods) influence the performance of the filter approach, so it makes sense to be conservative. I can also imagine that many data scientists will want to use different imputation methods than what are available in SimpleImputer.

manojbalaji1:
@t-tte in that case, we can raise error as mentioned and provide null_impute as an optional variable if they want to use imputation, or skip. This way they have an option to impute by themselves and then provide the dataframe, skip or use the options mentioned in this PR(the one that uses simpkeimputer). What do you say?

t-tte:
This sounds good to me @manojbalaji1 👍

zhenyuz0500:
Thanks for making this fix! 
Adding an error message and providing x_bin.nanmax() as an option to fix this issue sounds good to me. 
It might influence the performance of the function, depending on the proportion of missing value and the mechanism. Therefore, in the future, considering advanced imputation methods, such as creating another binary row indicating if a feature value is missing can be helpful in severe missing value scenario.

paullo0106:
Thanks for all the great discussions! Seems like we are aligned to merge this. For other potential enhancements, we can add them later in another PR.

PR #449, opened 2022-01-08
One typo in About

JackRab:
There is an extra space in `messaging`


paullo0106 (approved):
Thank you for the correction!

PR #476, opened 2022-02-13
Merely define preferred backend for joblib instead of hard-coding it

heiderich:
## Proposed changes

This fixes #475 by merely indicating a preference for the backend instead of hardcoding it. This way users can specify the backend using `joblib.parallel_backend` (see https://joblib.readthedocs.io/en/latest/parallel.html#parallel) according to their circumstances.

## Types of changes

What types of changes does your code introduce to **CausalML**?
_Put an `x` in the boxes that apply_

- [x] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Update (if none of the other choices apply)

## Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._

- [x] I have read the [CONTRIBUTING](https://github.com/uber/causalml/blob/master/CONTRIBUTING.md) doc
- [x] I have signed the [CLA](https://cla-assistant.io/uber/causalml)
- [x] Lint and unit tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [ ] Any dependent changes have been merged and published in downstream modules

## Further comments

Related: #468

PR #477, opened 2022-02-13
Allow parallel prediction for UpliftRandomForestClassifier

heiderich:
## Proposed changes

Previous to this PR `UpliftRandomForestClassifier` was using `joblib` to fit the `UpliftTreeClassifier`s in parallel if `n_jobs` was not equal to 1. With this PR predictions are also computed in parallel (one job per tree).

## Types of changes

## Checklist


causalml/inference/tree/uplift.pyx
```diff
@@ -1366,7 +1366,14 @@ class UpliftRandomForestClassifier:
 
         '''
         # Make predictions with all trees and take the average
-        y_pred_ensemble = sum([tree.predict(X=X) for tree in self.uplift_forest]) / len(self.uplift_forest)
+
+        if self.n_jobs != 1:
+            y_pred_ensemble = sum(
+                Parallel(n_jobs=self.n_jobs, prefer="threads")
```
jeongyoonlee, line 1372: Let's make the `joblib` backend as an optional input argument with `"threads"` as default to `__init__()` so that we can use e.g. `self.parallel_backend` in `fit()` and `predict()` instead of the hardcoded string.

jeongyoonlee (changes requested):
Thanks @heiderich! Looks good to me. I left one minor change request.

causalml/inference/tree/uplift.pyx
heiderich: I completely agree. I propose to do this once #476 is merged. This way the new argument can be used both for `predict` and `fit`.

jeongyoonlee: Sounds good! Just merged #476. Thanks!

causalml/inference/tree/uplift.pyx
```diff
@@ -1256,7 +1260,8 @@ class UpliftRandomForestClassifier:
                  n_reg=10,
                  evaluationFunction='KL',
                  normalization=True,
-                 n_jobs=-1):
+                 n_jobs=-1,
+                 joblib_prefer: str = "threads"):
```
jeongyoonlee, line 1264: nit: We're going to use [`black`](https://black.readthedocs.io/en/stable/) as a formatter as discussed in #474. You can use it integrated in your IDE or run its command line tool to ensure that the code style follows [PEP-8](https://www.python.org/dev/peps/pep-0008/#other-recommendations). e.g. no white space around `=` when used in input arguments.

Since we already have a separate PR, you don't need to update the style here.

jeongyoonlee (approved):
LGTM. Thanks!

Issue #498, opened 2022-04-16
RLearner cannot take effect_learner which does not have 'sample_weight' in the fit() argument


zhenyuz0500:
**Describe the bug**
Currently RLearner can only take effect_learner which has  'sample_weight' in the fit() argument, such as RandomForestRegressor() in sklearn. 
However, some regressor may not have 'sample_weight' in the fit() argument, such as MLPRegressor() in sklearn. Fitting such regressor in the RLearner will raise an error.

**To Reproduce**
MLPR = MLPRegressor(
    hidden_layer_sizes=(100,10,),
    max_iter=100,
    alpha=1e-4,
    solver="adam",
    verbose=False,
    random_state=1,
    learning_rate='adaptive'
)
rlearner=BaseRClassifier(outcome_learner = MLPC, control_name=control_group,effect_learner = MLPR)
rlearner.fit(x,t,y)

**Expected behavior**
The Rlearner should be flexible about the regressor.
We can first check if 'sample_weight' is in the fit() function, and proceed accordingly.

**Screenshots**

**Environment (please complete the following information):**
 - OS: [e.g. macOS, Windows, Ubuntu]
 - Python Version: [e.g. 3.6, 3.7]
 - Versions of Major Dependencies (`pandas`, `scikit-learn`, `cython`): [e.g. `pandas==0.25`, `scikit-learn==0.22`, `cython==0.28`]

**Additional context**

PR #500, opened 2022-04-22
Issue #498: Fixing sample_weight requirement in the Rlearner

zhenyuz0500:
## Proposed changes

This PR resolves issue #498 .

Making sample_weight an optional argument for base regressor learner for Rlearner:
1) if there is a sample_weight is in the base learner (such as random forest regressor), then proceed as usual (same previous version);
2) if the base learner does not take sample_weight (such as neural network), then skip this argument;

## Types of changes

- [x ] Bugfix (non-breaking change which fixes an issue)

## Checklist

- [x ] I have read the [CONTRIBUTING](https://github.com/uber/causalml/blob/master/CONTRIBUTING.md) doc
- [x ] Lint and unit tests pass locally with my changes
- [x ] I have added tests that prove my fix is effective or that my feature works

## Further comments

### DIFF
--- a/causalml/inference/meta/rlearner.py
+++ b/causalml/inference/meta/rlearner.py
@@ -1,3 +1,4 @@
+import inspect
 from copy import deepcopy
 import logging
 import numpy as np
@@ -488,9 +489,14 @@ class BaseRClassifier(BaseRLearner):
                         group
                     )
                 )
-            self.models_tau[group].fit(
-                X_filt, (y_filt - yhat_filt) / (w - p_filt), sample_weight=weight
-            )
+            if 'sample_weight' in inspect.getargspec(self.models_tau[group].fit).args:
+                self.models_tau[group].fit(
+                    X_filt, (y_filt - yhat_filt) / (w - p_filt), sample_weight=weight
+                )
+            else:
+                self.models_tau[group].fit(
+                    X_filt, (y_filt - yhat_filt) / (w - p_filt)
+                )                
 
     def predict(self, X, p=None):
         """Predict treatment effects.


paullo0106 (commented):
Looks good to me, thanks @zhenyuz0500 ! 

I noticed the build error [here](https://github.com/uber/causalml/runs/6133004316?check_suite_focus=true) and was trying to figure it out. Did some quick search and not sure if it's due to the [DeprecationWarning](https://stackoverflow.com/questions/61933427/deprecationwarning-inspect-getargspec-is-deprecated-since-python-3-0) of `inspect.getargspec()` during pytest stage and could replacing it with `inspect.getfullargspec()` help ([source](https://github.com/pytorch/pytorch/issues/15344))

t-tte:
I'm worried that not using sample weights will mess up the R-Learner results. Should we at least raise a warning?

In the longer term, it would be beneficial to calculate the sample weight automatically for methods that do not accept it as an argument. The challenge is that, based on a quick search, different methods use the sample weights slightly differently.

One simple approach is to create a new synthetic sample that has observations proportionally to the sample weights. This synthetic data set could be passed as a training set to any learner.

zhenyuz0500:
> I'm worried that not using sample weights will mess up the R-Learner results. Should we at least raise a warning?
> 
> In the longer term, it would be beneficial to calculate the sample weight automatically for methods that do not accept it as an argument. The challenge is that, based on a quick search, different methods use the sample weights slightly differently.
> 
> One simple approach is to create a new synthetic sample that has observations proportionally to the sample weights. This synthetic data set could be passed as a training set to any learner.

After reviewing the R Learner method, I realized the sample_weight is a required argument as R Learner is based on weighted regression - this is different from other learners. The only case where it is not needed is when the sample is equally split between treatment and control, as the weight will be the same (W_i-e*(X_i))^2 = 1/4.

Therefore, it may not be a good idea for merging this PR unless the user really understands what they are doing.

PR #539, opened 2022-08-24
Bart

caioguirado:
## Proposed changes

This PR proposes the implementation of Bayesian Additive Regression Trees (BART) as an additional method to the package.
The implementation allows the usage of BART for both a Classic ML problem setting and Uplift Modeling. The reson for including also the Classic ML setting was to allow easier validation of the method with synthetic data.

Currently the method works for regression and binary classification response types, and with binary treatment type.

References:
[[1] Chipman et al. (2010)](https://arxiv.org/abs/0806.3286)\
[[2] Hill (2011)](https://www.researchgate.net/publication/236588890_Bayesian_Nonparametric_Modeling_for_Causal_Inference)\
[[3] Kapelner and Bleich (2014)](https://arxiv.org/abs/1312.2171)\
[[4] Tan and Roy (2019)](https://arxiv.org/abs/1901.07504)\
[BartPy](https://github.com/JakeColtman/bartpy)

## Types of changes

- [x] Documentation Update (if none of the other choices apply)

## Checklist

## Further comments

Some next steps are proposed for improvement:
- Add parallelization:
    In the example notebook added, there's a `cProfile` analysis of the methods that take the most time to execute inside the `fit` method. Both the computation of the individual tree residuals and prediction step are the top opportunities of improvement. They also have a very similar logic. [Pratola et al.](https://arxiv.org/abs/1309.1906) proposed a way of parallelizing it.
    
- Add non-binary treatment support.
- Add multi-class classification support.
- Add MCMC statistics report and confidence intervals for BART predictions


jeongyoonlee:
Thanks for the PR for BART, @caioguirado! 

I will take a look further in details, but have two comments at the moments as follows:
- [ ] Currently, it's failing at [the Lint test](https://github.com/uber/causalml/runs/8020453870?check_suite_focus=true). Could you please run `black` to reformat the code?
- [ ] The test data set used in the example notebook looks too small. Could you please use more features e.g. ~10 with the mix of informative and non-informative features?

Thanks!

causalml/inference/tree/bart.py
```diff
@@ -0,0 +1,547 @@
+import copy
+import random
+import numpy as np
+import pandas as pd
+from tqdm import tqdm
+import scipy.stats as ss
+import statsmodels.api as sm
```
t-tte, line 7: Looks like we can drop this line.

causalml/inference/tree/bart.py
```diff
@@ -0,0 +1,547 @@
...
+    
+    def update_residuals(self, residuals:np.ndarray) -> None:
+        self.root.data.R = residuals
+        self.traverse(self.root, type_='update_residuals')
+        
+    def traverse(self, root:Node, type_:str = 'update_leaves'):
+        if type_ == 'update_leaves':
+            if not root.left and not root.right:
+                self.leaves.append(root)
+                return
```
t-tte, line 91: It's not clear to me why we need to return None here (and below).

causalml/inference/tree/bart.py
```diff
@@ -0,0 +1,547 @@
...
+            if self.task == 'regression':
+                response = transformed_y
+            else:
+                sample = np.random.normal(loc=self.sum_of_trees, scale=1, size=(len(transformed_y), 1))
+                max_els = self.max_el(sample)
+                min_els = self.min_el(sample)
+                response = np.where(y.reshape(-1, 1) == 1, max_els, min_els) # Z in the paper
+
+            for j in range(len(self.trees)):
+                residuals = self.compute_residual(X=X, y=response, trees=self.trees, j=j) # opportunity to parallelize according to Pratola et al. (2013) (https://arxiv.org/abs/1309.1906)
```
t-tte, line 477: Not sure what it says in the reference, but if I understand things correctly, for each tree we compute the residuals by looping over every other tree, which indeed seems wasteful. Is it not possible to store the prediction of each tree to avoid computing the same prediction over and over again?

causalml/inference/tree/bart.py
```diff
@@ -0,0 +1,547 @@
...
+            predictions = []
+            for treatment in self.treatments:
+                treatment_predictions = self._predict(np.concatenate([X, np.ones((len(X), 1)) * treatment], axis=1), rescale)
+                predictions.append(treatment_predictions.flatten().tolist())
+                columns.append(f'treatment_{treatment}' if treatment > 0 else 'control')
+            
+            df_res = pd.DataFrame(np.array(predictions).T, columns=columns)
+            
+            # From: https://github.com/uber/causalml/blob/c42e873061eb74ec9c3ca6ea991e113b886245ae/causalml/inference/tree/uplift.pyx
+            df_res['recommended_treatment'] = df_res.apply(np.argmax, axis=1)
```
t-tte, line 532: Not necessarily something to solve here, but we need to change the terminology of "recommended treatment", because CATE is just one factor that is relevant for whether a treatment should be recommended or not.

causalml/inference/tree/bart.py
t-tte: My bad, I saw OLS being called later on.

t-tte:
Dropped some comments but overall looks great to me. The example notebook fails to render for me at the moment but I'll take a look if/when you've added more predictors as Jeong suggested.

zhenyuz0500:
One of the BART features is supporting continuous treatment - it will be an excellent add to the package if this is supported. I'm wondering how much effort is needed to support continuous treatment. 

Other than that, it will be great to compare BART's performance with some other models and show its advantage and value. Or maybe mention BART can support which scenarios that cannot be solved by other models in the current implementation.

ras44:
@jeongyoonlee happy to take a look at this PR again too

aman-coder03:
Hi @jeongyoonlee 
I came across this PR while looking through open contributions and noticed that it has been inactive for some time
I'd be interested in helping revive this BART implementation if there is still interest in adding it to the library.
Before I start digging into the merge conflicts and review comments, could the maintainers share whether BART is still considered a desirable addition to CausalML?
If so, I'd be happy to investigate the current state of the branch, address outstanding review feedback, and work towards getting it compatible with the current codebase

Gattocrucco:
Somehow I ran across this PR. This PR implements BART in pure Python from scratch, but I want to signal that in 2026 the state of the BART ecosystem in Python is much improved and there are two packages that could be pulled in as dependencies that implement BART much more efficiently: https://github.com/bartz-org/bartz (by me) and https://github.com/StochasticTree/stochtree .

aman-coder03:
Hi @jeongyoonlee , i am following up on my earlier message and also flagging @Gattocrucco 's recent comment
before I start rebasing the original implementation and addressing the outstanding review feedback, I wanted to get maintainer input on direction, specifically whether you would prefer...

1. **Rebase the original implementation**
resolve merge conflicts, run `black`, fix the residual caching issue raised by @t-tte, expand the example notebook, and address the other review comments. Straightforward but we'd be maintaining a pure-Python BART from scratch

2. **Wrap `bartz` or `stochtree` as a dependency**
as @Gattocrucco suggests, both are mature, well-optimized implementations. This would likely give better performance and lower maintenance burden, but is a different scope of work

what's your preference?

jeongyoonlee:
Thanks, @Gattocrucco and @aman-coder03. Given the current state of the new BART Python ecosystem, it makes sense to discard this PR and instead add bartz (or an alternative) as a dependency. Closing this.

Issue #534, opened 2022-08-11
Problem with installing the causalml library


mwasielewski15:
**Describe the bug**
 I am trying to install causalml package both using pip and from source following these instructions: https://causalml.readthedocs.io/en/latest/installation.html
Unfortunately I cannot succeed with building wheel for causalml (screenshot below).
It looks like there is a problem with setup.py file which is part of repository I had to clone (according to the installation instructions).
I've also installed Microsoft Visual C++ as it's suggested in error but it didn't change anything.

**To Reproduce**
$ git clone https://github.com/uber/causalml.git
$ cd causalml
$ pip install -r requirements.txt
$ pip install causalml  <- after this command I get an error

**Expected behavior**

**Screenshots**
![image](https://user-images.githubusercontent.com/75525959/184136410-755f4a16-cae3-4cbb-89d8-91c0ab0e86e8.png)

**Environment (please complete the following information):**
 - Windows
 - Python 3.10
 - Versions of Major Dependencies : according to requirements.txt from repository https://github.com/uber/causalml.git

alexander-pv:
Hi, in WIndows OS case check that:
-  MVC++ is in Path environment variable
-  80 lines of further output doesn't contain errors of finding C/C++ headers

bsaunders27:
FYI I'm facing the same issues on a different system, tested building with 0.13.0 and current 0.14.0 but both are breaking with different errors.

**Environment**
- MacOS Ventura 13.4.1 - M1
- Python 3.8/3.9 (both installed with `conda env create -f environment-py3*.yml`)
- Versions of major dependencies - See above

```
Processing /Users/benjaminsaunders/Documents/Marketing/causalml
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  error: subprocess-exited-with-error
  
  × Getting requirements to build wheel did not run successfully.
  │ exit code: 1
  ╰─> [62 lines of output]
      warning: /private/var/folders/ck/8kpq2yv50h18k9z3jwktl1gm0000gp/T/pip-build-env-1tpq9b41/overlay/lib/python3.8/site-packages/sklearn/tree/_tree.pxd:61:73: The keyword 'nogil' should appear at the end of the function signature line. Placing it before 'except' or 'noexcept' will be disallowed in a future version of Cython.
[...]

PR #641, opened 2023-07-18
Fix for builds on Mac M1 infrastructure

bsaunders27:
## Proposed changes

Following up with issues I saw with pip install in M1 architecture. Deep dived the issue trying `conda env create` and found issues there as well, so added some fixes for that build on M1 (mostly updating for compatability with `numpy` for Darwin arm64 sys build). Breaking out as separate commits to simplify cherry picking since some issues may not be generalizable. Known related issue https://github.com/uber/causalml/issues/534. 

## Types of changes

## Checklist

## Further comments

None, other than the fact that this is my first PR and I would not be surprised if there's a more elegant resolution

### DIFF
--- a/envs/environment-py39.yml
+++ b/envs/environment-py39.yml
@@ -2,7 +2,7 @@ name: causalml-py39
 channels:
   - defaults
 dependencies:
-  - pip=21.0.1
+  - pip
   - python=3.9
   - setuptools=52.0.0
   - pip:
--- a/envs/environment-py39.yml
+++ b/envs/environment-py39.yml
@@ -5,6 +5,7 @@ dependencies:
   - pip
   - python=3.9
   - setuptools=52.0.0
+  - numpy=1.22
   - pip:
     - absl-py==0.12.0
     - appnope==0.1.2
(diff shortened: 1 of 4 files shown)


nelehuvaere:
The change in _builder.pyx also solves the cython compiling error on windows. Thanks for the pr, hope it gets merged.

jeongyoonlee (approved):
LGTM. Thanks for the contribution!

bsaunders27:
ah fiddle dee dee, let me see what broke

envs/environment-py39.yml
```diff
@@ -5,6 +5,7 @@ dependencies:
   - pip
   - python=3.9
   - setuptools=52.0.0
+  - numpy=1.22
```
eugeneyarovoi, line 8: Curious to know, what is the difference between putting numpy in this section, vs. the pip section? You don't want to specify a minor version like for all the other dependencies?

jeongyoonlee: The dependencies listed above will be installed using conda, while the ones in the pip section will use pip. If available, conda installation is more robust as it packages binaries and executables necessary for installation together with the Python library, but not all packages are available in conda.

bsaunders27: Hey @eugeneyarovoi , so initially I'd added above to conda install since numpy was required for pip wheel building prior to installs and saw failures for anything before `numpy==1.22.0`. I tested again recently and after updating all of the subsequent package versions and looks like it's fine, but if there's a preference for conda as the package handler re: @jeongyoonlee can leave as is. Did find a few other items breaking that I'd like to update quick for the arm64x build on mac so might have a quick fast-follow

PR #656, opened 2023-08-14
fix: simulate_randomized_trial

bsaunders23:
## Proposed changes

Caught a bug in the simulation of randomized trial data, prior code of `np.maximum(np.repeat(0.0, n), X[:, 0] + X[:, 1], X[:, 2])` was actually returning `np.maximum(np.repeat(0.0,n), X[:,0] + X[:,1])` and assigning the value of this to X[:,2] based on the documentation (see [out](https://numpy.org/doc/stable/reference/generated/numpy.maximum.html)).

## Types of changes

## Checklist

## Further comments

Impact of this is issues with parameter recall on different models (X[:,2] suddenly shows up as super important under shap values when it shouldn't be related to tau). Illustration of data impact included below:

image

### DIFF
--- a/causalml/dataset/regression.py
+++ b/causalml/dataset/regression.py
@@ -102,7 +102,7 @@ def simulate_randomized_trial(n=1000, p=5, sigma=1.0, adj=0.0):
     """
 
     X = np.random.normal(size=n * p).reshape((n, -1))
-    b = np.maximum(np.repeat(0.0, n), X[:, 0] + X[:, 1], X[:, 2]) + np.maximum(
+    b = np.maximum.reduce([np.repeat(0.0, n), X[:, 0] + X[:, 1], X[:, 2]]) + np.maximum(
         np.repeat(0.0, n), X[:, 3] + X[:, 4]
     )
     e = np.repeat(0.5, n)


causalml/dataset/regression.py
```diff
@@ -102,7 +102,7 @@ def simulate_randomized_trial(n=1000, p=5, sigma=1.0, adj=0.0):
     """
 
     X = np.random.normal(size=n * p).reshape((n, -1))
-    b = np.maximum(np.repeat(0.0, n), X[:, 0] + X[:, 1], X[:, 2]) + np.maximum(
+    b = np.maximum.reduce([np.repeat(0.0, n), X[:, 0] + X[:, 1], X[:, 2]]) + np.maximum(
         np.repeat(0.0, n), X[:, 3] + X[:, 4]
```
jeongyoonlee, line 106: Can you add `assert p >= 5, "p, the number of covariates, should be equal or greater 5."`?

jeongyoonlee: Actually, there's the same pattern in other DGP functions as well, and should be addressed in a separate PR altogether.

jeongyoonlee (approved):

Issue #662, opened 2023-08-23
Add multi-armed Qini (maq)


t-tte:
**Is your feature request related to a problem? Please describe.**
Current methods for evaluating multiple costly treatments are somewhat hacky.

**Describe the solution you'd like**
The good folks at [grf labs](https://github.com/grf-labs/maq) came up with a generalisation of the Qini curve method for multiple costly treatments.

**Describe alternatives you've considered**
N/A

**Additional context**
@erikcs is currently working to polish the Python wrapper for maq. The idea is to incorporate this in a lightweight way to Causal ML.

ras44:
hi @t-tte

FYI, looks like the MAQ python package is available now: https://github.com/grf-labs/maq/tree/master/python-package

with example: https://github.com/grf-labs/maq/blob/master/python-package/notebooks/introduction.ipynb

t-tte:
Yep, @erikcs has now now polished the Python implementation.

The idea we discussed was to implement multi-qini through CausalML.metrics.multiqini or something similar. Should be doable via a lightweight wrapper. Does this seem reasonable @erikcs, @ras44 and @jeongyoonlee?

I'm currently short on time but can look into this early next year. Alternatively, if any of you have bandwidth for an initial implementation, I'll be happy to review.

erikcs:
Thanks @t-tte, that sounds great! I too have a bit limited bandwidth now, but would be happy to provide input (and could also gladly hop on a zoom if anyone wants a quick live tour of how the multi-armed qini proposal works). And nice to see you @ras44, I remember you made some nice contributions to grf!

jeongyoonlee:
I'm closing this issue as we merged #729.

PR #729, opened 2023-12-11
minimal wrapper of MAQ #662

ras44:
## Proposed changes

Added a minimal wrapper for MAQ, a simple test, and dependency on the github repo v0.2.0.

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Checklist

- [X] I have read the [CONTRIBUTING](https://github.com/uber/causalml/blob/master/CONTRIBUTING.md) doc
- [X] I have signed the [CLA](https://cla-assistant.io/uber/causalml)
- [X] Lint and unit tests pass locally with my changes
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] I have added necessary documentation (if appropriate)
- [X] Any dependent changes have been merged and published in downstream modules

## Further comments

N/A


causalml/metrics/multiqini.py
```diff
@@ -0,0 +1,17 @@
+from maq import MAQ, get_ipw_scores
+
+
+def get_mq_ipw_scores(Y, W, W_hat=None):
+    """Construct evaluation scores via inverse-propensity weighting.
```
erikcs, line 5: Is possible to programmatically have this function inherit the docstring from get_ipw_scores? (and similarly below, does MultiQini get MAQ's docstring?)

erikcs:
This looks like a very clean and simple integration. Before you lock onto the 0.2.0 tag: I'm happy to make changes downstream if there's anything there you think could be improved.

Also, does copy/pasting the [maq intro notebook](https://github.com/grf-labs/maq/tree/master/python-package/notebooks) into the docs here make sense?

causalml/metrics/multiqini.py
ras44: hi @erikcs please see commit below that should add that 👍

ras44:
thanks @erikcs- appreciate the contributions!

Deferring to @t-tte and @jeongyoonlee for comments:
- implementation to inherit docstrings from MAQ
- any upstream change requests for MAQ before creating a version dependency?
- incorporating [maq intro notebook](https://github.com/grf-labs/maq/tree/master/python-package/notebooks)

ras44:
@erikcs if it's ok with you, I thought I might rename the wrapper for `get_ipw_scores` to something like `get_ipw_eval_scores` to follow the terminology in the [paper](https://arxiv.org/pdf/2306.11979.pdf).  The reason is that there is a function in CausalML called `get_simple_iptw` which returns IPTW weights.  So `get_ipw_scores` could possibly be misinterpreted as returning the per-arm IPTW weights instead of the per-arm ipw-weighted eval scores.

erikcs:
Thanks @ras44, I agree, `get_ipw_eval_scores` sounds more informative!  I'd like to make that function rename downstream in the python wrapper too, I think having causalml and the original python wrapper use consistent naming could help reduce potential confusion in the future. On that note, maybe we can decide on `MultiQini` vs `MAQ` too?  I think a pro of `MAQ` is that it emphasizes that it is a different animal from "Qini", it's a Qini for many arms, `MultiQini` might give the impression that it is just many Qini curves?

ras44:
Good point- I think @t-tte had suggested MultiQini, so perhaps he can provide some input to norm on a naming convention?  happy to follow your lead

t-tte:
Yep, MAQ sounds great to me. I think it's also a good idea to copy paste the notebook. As regards inheriting the doc strings, if it's technically feasible, I don't think anyone has a philosophical objection against it.

erikcs:
@ras44 I don't have push access to your branch, but if you add 
```Python
from .multiqini import MAQ, get_ipw_scores # noqa
```
to `metrics/__init__.py`

Then CausaML's documentation builds what you would expect, at least on my laptop: (using maq tag `py0.2.2`)
Screen Shot 2023-12-23 at 20 14 56

I think it's fine just doing 
```Python
 from maq import MAQ, get_ipw_scores
```
in `metrics/multiqini.py` since `get_ipw_scores` lives in another module from `get_simple_iptw`?

If I copy the [notebook](https://github.com/grf-labs/maq/blob/master/python-package/notebooks/introduction.ipynb) into `docs/examples/qini_curves_for_costly_treatment_arms.ipynb` and just change the line `from maq import MAQ, get_ipw_scores` to `from causalml.metrics.multiqini import MAQ, get_ipw_scores` then that too works fine here.

ras44:
hi @erikcs Thanks for your comments- just getting back to this!

If we opt not to create a wrapper class called MultiQini in `multiqini.py` (and it does sound like we've decided on following MAQ naming conventions), then could we just add:

```
from maq import MAQ, get_ipw_scores # noqa
```

in `metrics/__init__.py`?

That builds the CausalML documentation as expected on my machine, and avoids creating the additional `multiqini.py` file entirely.

I haven't tested the notebook yet, but I think the above setup should be able to run the example notebook as you've described, i.e. using `from causalml.metrics import MAQ, get_ipw_scores`.

Finally, re xgboost in the doc build: xgboost is a dependency, so I think it should be available during the build if I'm understanding correctly:

https://github.com/uber/causalml/blob/06a5f8fb720f4f3157322518f115c8d53983382d/pyproject.toml#L37


If that sounds good, I'll add the above import to `metrics/__initi__.py` and remove the `multiqini.py` file entirely.

@jeongyoonlee @t-tte

erikcs:
Thanks @ras44, that sounds great here. I guess another final thing is how to best integrate maq via pyproject.toml, since the package installs from github, a user now needs git and a c++ compiler to install causalml, if that is problematic, it could maybe be added as an optional dependency instead of a dependency?

ras44:
@erikcs just added a couple commits and the MAQ import should be all set.  The tests are also passing.  Last step is probably getting that notebook in.

Re your comments about needing git and a c++ compiler: I think we assume git is installed, but our installation instructions previously included a recommendation and details on how to set up a conda environment with a variety of required system libraries/executables (including a c++ compiler via `cxx-compiler`):

https://github.com/uber/causalml/blob/028c63cfccd23b448b526d965864c316e95d1c7c/README.md?plain=1#L112-L120

In a recent rewrite of `README.md` we intended to move these instructions into the installation section of the docs, but the above seems to have been missed, so I'll fix that in a new PR.  That should ensure that anyone following the instructions has the proper environment set up for the installation/build of MAQ too.

pyproject.toml
```diff
@@ -45,6 +45,7 @@ dependencies = [
     "torch",
     "pyro-ppl",
     "graphviz",
+    "maq@git+https://github.com/grf-labs/maq.git@py0.2.0#egg=maq&subdirectory=python-package",
 ]
```
erikcs, line 49: I tried to make some documentation clarifications that are available if you change this to: `@py0.2.2`

erikcs:
Thank you @ras44!

ras44:
@jeongyoonlee @erikcs @t-tte I think we're set to merge, would appreciate any final thoughts!

jeongyoonlee (approved):
LGTM. Thanks, @ras44 and @erikcs, for the contribution.

PR #790, opened 2024-09-13
Make torch optional. Update DragonNet w/ latest TF APIs

### LINKED ISSUE #789: Make torch an optional dependency

The `torch` package, which is currently a required dependency for `causalml`, requires presence of CUDA libraries like `libcublas.so.*[0-9]`. This causes installation issues, especially in environments where these libraries are not pre-installed or compatible, leading to errors such as `ValueError: libcublas.so.*[0-9] not found in the system path`. However, within `causalml`, `torch` is [only used for the Causal Effect VAE model](https://github.com/search?q=repo%3Auber%2Fcausalml%20torch&type=code), which relies on neural networks. 

`torch` is also a large package, would be good to not have to install it by default.

To resolve this, I propose making `torch` an optional dependency within `causalml` via `pip install causalml[torch]`. This would be similar to the previously implemented approach with TensorFlow in https://github.com/uber/causalml/pull/343, where it was made optional and only installed when needed via `pip install causalml[tf]`. 

* A workaround for envs with no GPUs is to specify the following in your Pipfile when you want to add `causalml`: 
```
[[source]]
name = "pytorch"
url = "https://download.pytorch.org/whl/cpu"
verify_ssl = true

torch = {version = "==2.1.2", index = "pytorch"}
```
* A better, but way more involved, alternative would be to separate training deps from prediction deps as suggested in: https://github.com/uber/causalml/issues/594

jeongyoonlee:
## Proposed changes

This PR fixes #789 by making the torch dependency optional.

It also fixes the `DragonNet` test error with the latest Tensorflow APIs.
 
## Types of changes

## Checklist

## Further comments

N/A


tests/test_dragonnet.py
```diff
@@ -3,21 +3,21 @@
...
-def test_save_load_dragonnet():
+def test_save_load_dragonnet(tmp_path):
     y, X, w, tau, b, e = simulate_nuisance_and_easy_treatment(n=1000)
 
     dragon = DragonNet(neurons_per_layer=200, targeted_reg=True, verbose=False)
     dragon_ite = dragon.fit_predict(X, w, y, return_components=False)
     dragon_ate = dragon_ite.mean()
-    dragon.save("smaug")
+
+    model_file = tmp_path / "smaug.h5"
```
paullo0106, line 17: @jeongyoonlee did you mean `f"{tmp_path}/smaug.h5"` ? I'm not super familiar with the code here

jeongyoonlee: tmp_path is a Pathlib object and this is a valid syntax. :)

paullo0106 (approved):
Thanks!

tests/test_dragonnet.py
paullo0106: i see, makes sense then!

PR #817, opened 2025-02-22
resolves #811, update propensity.py

### LINKED ISSUE #811: modifications to propensity.py

A variety of topics came up re `propensity.py` in the last CausalML call:
 
- `compute_propensity_score` could either:
  - enforce that the `p_model` arg is an instance of the PropensityModel class
    - if so, then alter the Propensity model class `predict` and `predict_proba` functions so that `predict` calls `predict_proba` if the underlying model doesn't have a `predict` function that produces scores (and not classifications)
  - accept any model passed (for instance, a Naive Bayes classifier)
    - if so, then call `predict_proba` or `predict` based on the availability of `predict_proba` from within `compute_propensity_score` for the passed model, so as to always produce the propensity scores and not the classification

- when calling `compute_propensity_score` with `calibrate=True`, the model that is returned is not the model that produces the returned propensity scores (those are calculated by calibration function model).  This seems like it could lead to confusion if someone were to attempt to reproduce the scores using the returned model.
  - Would it make sense here to return `None` for the model? Or some other warning if `calibrate=True`?

Not sure about the tradeoffs between accepting any model in `compute_propensity_score` vs requiring that the model be an instance of `PropensityModel`, but I'd be inclined to accept any model (i.e. a Naive Bayes classifier as was used in the calibration example) and then test for the existence of `predict_proba` to produce scores, before using `predict` if `predict_proba` doesn't exist.  Something like:

```
def compute_propensity_score(
    X, treatment, p_model=None, X_pred=None, treatment_pred=None, calibrate_p="iso"
):
    """Generate propensity score if user didn't provide

    Args:
        X (np.matrix): features for training
        treatment (np.array or pd.Series): a treatment vector for training
        p_model (propensity model object, optional):
            ElasticNetPropensityModel (default) / GradientBoostedPropensityModel
        X_pred (np.matrix, optional): features for prediction
        treatment_pred (np.array or pd.Series, optional): a treatment vector for prediciton
        calibrate_p (bool, optional): whether calibrate the propensity score

    Returns:
        (tuple)
            - p (numpy.ndarray): propensity score
            - p_model (PropensityModel): a trained PropensityModel object
    """

    print("using local compute_propensity_score")
    
    if treatment_pred is None:
        treatment_pred = treatment.copy()
    if p_model is None:
        p_model = ElasticNetPropensityModel()

    p_model.fit(X, treatment)

    if X_pred is None:
        try:
            p = p_model.predict_proba(X)[:, 1]
        except AttributeError:
            print("predict_proba not available, using predict instead")            
            p = p_model.predict(X)        
    else:
        try:
            p = p_model.predict_proba(X_pred)[:, 1]
        except AttributeError:
            print("predict_proba not available, using predict instead")            
            p = p_model.predict(X_pred)

    if calibrate_p:
        print("Isotonic calibrating propensity scores only.  Returning model=None.")
        p = calibrate_iso(p, treatment_pred)
        p_model = None

    # force the p values within the range
    eps = np.finfo(float).eps
    p = np.where(p < 0 + eps, 0 + eps * 1.001, p)
    p = np.where(p > 1 - eps, 1 - eps * 1.001, p)

    return p, p_model

```
 

Refs:

https://github.com/uber/causalml/blob/22b13c2594a491317908feb54c9f3401bafd9d43/causalml/propensity.py#L14

https://github.com/uber/causalml/blob/22b13c2594a491317908feb54c9f3401bafd9d43/causalml/propensity.py#L201

ras44:
## Proposed changes

In `propensity.py` modify the `compute_propensity_score(...)` function to:

- Try predict_proba and fall back to predict for any classifier.
- Return p_model=None if calibrate_p=True.

## Types of changes

## Checklist

## Further comments

This is one potential update for `compute_propensity_score(...)`.  We'd also talked about modifying the `PropensityModel` class itself, however this would not allow for the passing of an arbitrary classifier like Naive-Bayes via the `p_model` argument to `compute_propensity_score`.  Nevertheless, updating `PropensityModel::predict(...)` and `PropensityModel::fit_predict(...)` might still also make sense for further robustness.

### DIFF
--- a/causalml/propensity.py
+++ b/causalml/propensity.py
@@ -223,13 +223,13 @@ def compute_propensity_score(
 
     p_model.fit(X, treatment)
 
-    X = X if X_pred is None else X_pred
+    X_pred = X if X_pred is None else X_pred
 
     try:
-        p = p_model.predict_proba(X)[:, 1]
+        p = p_model.predict_proba(X_pred)[:, 1]
     except AttributeError:
         logger.info("predict_proba not available, using predict instead")
-        p = p_model.predict(X)
+        p = p_model.predict(X_pred)
 
     if calibrate_p:
         logger.info("Calibrating propensity scores. Returning p_model=None.")
--- a/causalml/propensity.py
+++ b/causalml/propensity.py
@@ -201,13 +201,12 @@ def calibrate(ps, treatment):
 def compute_propensity_score(
     X, treatment, p_model=None, X_pred=None, treatment_pred=None, calibrate_p=True
 ):
-    """Generate propensity score if user didn't provide
+    """Generate propensity score if user didn't provide and optionally calibrate.
 
     Args:
         X (np.matrix): features for training
         treatment (np.array or pd.Series): a treatment vector for training
-        p_model (propensity model object, optional):
-            ElasticNetPropensityModel (default) / GradientBoostedPropensityModel
+        p_model (model object, optional): a binary classifier with either a predict_proba or predict method
         X_pred (np.matrix, optional): features for prediction
         treatment_pred (np.array or pd.Series, optional): a treatment vector for prediciton
         calibrate_p (bool, optional): whether calibrate the propensity score
@@ -215,7 +214,7 @@ def compute_propensity_score(
     Returns:
         (tuple)
             - p (numpy.ndarray): propensity score
-            - p_model (PropensityModel): a trained PropensityModel object
+            - p_model (PropensityModel): either the original p_model, a trained ElasticNetPropensityModel, or None if calibrate_p=True
     """
     if treatment_pred is None:
         treatment_pred = treatment.copy()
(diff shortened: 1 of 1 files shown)


causalml/propensity.py
```diff
@@ -224,14 +223,18 @@ def compute_propensity_score(
...
-        p = p_model.predict(X)
-    else:
-        p = p_model.predict(X_pred)
+    X0 = X if X_pred is None else X_pred
+
+    try:
+        p = p_model.predict_proba(X0)[:, 1]
+    except AttributeError:
+        logger.info("predict_proba not available, using predict instead")
+        p = p_model.predict(X0)
```
jeongyoonlee, line 232: Why don't we use `X` for `X_pred` if `X_pred` is none instead of creating a new variable `X0`?

jeongyoonlee: Sorry, @ras44. I meant using `X_pred` instead of `X`. 

The rationals are:
1. The code is for prediction, so it makes logical sense to use `X_pred` instead of `X`.
2. Modifying the input within a function other than the optimization purpose is not a good practice.

It's a nitpick and doesn't have any practical implication because, in this case, it won't change the original `X` and create a new variable only within the function, but, in general, there's a case you may unintentionally modify the input (e.g., updating values of numpy arrays or pandas dataframes with indexing).

jeongyoonlee (approved):
LGTM. Thanks!

causalml/propensity.py
ras44: absolutely- I should have caught that 😄

causalml/propensity.py
```diff
@@ -201,21 +201,20 @@ def calibrate(ps, treatment):
 def compute_propensity_score(
     X, treatment, p_model=None, X_pred=None, treatment_pred=None, calibrate_p=True
```
stasf25, line 202: Considering  [this issue](https://github.com/uber/causalml/issues/838#issue-3151477601), I'd propose to change the default value of the calibrate_p parameter to False

PR #834, opened 2025-05-27
install via `uv`

aaelony:
## Proposed changes

Simple illustration of how to install causalml using `uv`.

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -51,6 +51,35 @@ Install ``causalml`` with ``torch`` for ``CEVAE`` from ``PyPI``
     pip install causalml[torch]
 
 
+
+Install from ``PyPI`` using [`uv`](https://github.com/astral-sh/uv/blob/main/README.md)
+---------------------
+
+.. code-block:: bash
+
+    uv init myproject
+    uv venv
+    source .venv/bin/activate
+    uv pip install causalml
+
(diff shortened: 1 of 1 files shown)


jeongyoonlee (approved):
LGTM. Thanks for the contribution, @aaelony!

emilioMaddalena:
I think it's cool to have a section on `uv`, however, IMO the example added should be rewritten.

**Issue:** 
Having a section on `uv` but manually creating a venv and calling `pip` is not how `uv` is intended to be used.

**Details:** 
With `uv` one is not supposed to manage venvs manually, nor to use `pip` all the time. In fact, `uv` was intended to replace `pip` and the `uv pip install pkg` should be seen as a [temporary alternative](https://docs.astral.sh/uv/pip/compatibility/) for projects not yet ready to fully migrate.

**Proposed solution:** 
I would simplify the example to:
```python
uv init
uv add causalml
```
and the section title to `Install using uv` since the default index is PyPI anyway.

emilioMaddalena:
Hi @aaelony. Idk if you saw it, but I created a PR here #836.

Yeah, now that you mentioned it, I relized I have to fix my examples... The correct syntax is `uv add "causalml[tf]"`.

aaelony:
aha, quotes!  Makes sense :)

PR #850, opened 2025-09-24
Fix #848: Pass estimation_sample_size parameter to individual trees in UpliftRandomForestClassifier

### LINKED ISSUE #848: UpliftRandomForestClassifier: not passing estimation_sample_size to trees in fit method
`estimation_sample_size` is an input for UpliftRandomForestClassifier, but the value you pass in never gets passed to the trees in the forest, so it always defaults to 0.5 no matter what you input.

mohsinm-dev:
## Summary
Fixes issue #848 where the `estimation_sample_size` parameter was not being passed to individual trees in `UpliftRandomForestClassifier`.

## Changes
- Added parameter storage in `UpliftRandomForestClassifier.__init__` (line 2430)
- Added parameter passing to individual `UpliftTreeClassifier` instances during forest creation (line 2481)

## Root Cause
The constructor accepted the `estimation_sample_size` parameter but failed to:
1. Store it as an instance variable (`self.estimation_sample_size = estimation_sample_size`)
2. Pass it to individual `UpliftTreeClassifier` instances during forest creation

## Impact
This fix enables users to properly control honest splitting behavior in random forests when `honesty=True`. Previously, all trees defaulted to 0.5 regardless of user specification.

## Testing
- Verified existing test suite passes (102/103 tests pass, 1 unrelated GraphViz failure)
- Confirmed no breaking changes or regressions
- Tested with multiple `estimation_sample_size` values (0.1-0.9)
- Verified default behavior preservation (0.5 when not specified)

## Files Changed
- `causalml/inference/tree/uplift.pyx` (2 line addition)

### DIFF
--- a/causalml/inference/tree/uplift.pyx
+++ b/causalml/inference/tree/uplift.pyx
@@ -243,7 +243,7 @@ def group_uniqueCounts_to_arr(np.ndarray[TR_TYPE_t, ndim=1] treatment_idx,
         tv = treatment_idx[i]
         # assume treatment index is in range
         out_arr[2*tv] += 1
-        # assume y should be either 0 or 1, so this is summing 
+        # assume y should be either 0 or 1, so this is summing
         out_arr[2*tv + 1] += y[i]
     # adjust the entry at index 2*i to be N(Y = 0, T = i) = N(T = i) - N(Y = 1, T = i)
     for i in range(n_class):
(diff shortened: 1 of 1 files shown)


jeongyoonlee (approved):
LGTM. Thanks for your contribution, @mohsinm-dev!

PR #886, opened 2026-03-07
Add post-fit confidence intervals to `BaseTLearner` via `store_bootstraps` and `return_ci`

aman-coder03:
## Proposed changes
currently CIs are only available via `fit_predict(..., return_ci=True)`, which re-runsbootstrap resampling every time. This makes it impossible to get CIs on new data aftera model is already trained, which is a common need in production where you train once and score repeatedly

## How
added `store_bootstraps=False` to `fit()`. When set to True, it trains a bootstrap ensemble andstores it in `self.bootstrap_models_`. Added `return_ci=False` to `predict()`. When set to True,it scores the stored ensemble and returns `(te, te_lower, te_upper)` with no retraining needed
```python
learner.fit(X_train, treatment_train, y_train, store_bootstraps=True, n_bootstraps=200)
tau, lb, ub = learner.predict(X_test, return_ci=True)
```
all existing calls to `fit()`, `predict()` and `fit_predict()` are completely unchanged since the new parameters are opt-in only.

## Scope
this PR covers `BaseTLearner` as the reference implementation. I plan to extend the same pattern to the remaining learners once the design is confirmed by maintainers
Closes #885 

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/causalml/inference/meta/base.py
+++ b/causalml/inference/meta/base.py
@@ -78,6 +78,7 @@ class BaseLearner(metaclass=ABCMeta):
         X,
         treatment,
         y,
+        p=None,
         n_bootstraps=200,
         bootstrap_size=10000,
         random_state=None,
@@ -85,24 +86,29 @@ class BaseLearner(metaclass=ABCMeta):
     ):
         """Train and store a bootstrap ensemble for post-fit CI estimation.
 
-        Fits n_bootstraps cloned copies of the learner on bootstrap samples and
-        stores them in self.bootstrap_models_. Used by predict(return_ci=True)
-        to compute percentile-based confidence intervals on new data without
-        refitting.
+        Fits n_bootstraps cloned copies of the entire learner on bootstrap samples
+        and stores them in self.bootstrap_models_. Used by predict(return_ci=True)
+        to compute percentile-based confidence intervals on new data without refitting.
 
-        Note: storing N bootstraps of a GBM-based learner with k treatment groups
-        holds 2*N*k model objects in memory. Monitor RAM for large N or heavy
-        base learners.
+        This design follows EconML's BootstrapEstimator pattern — each bootstrap
+        clone is a full copy of the learner, making this method generic across all
+        meta-learners.
+
+        Note: storing N bootstrap clones can be memory-intensive for heavy base
+        learners. Monitor RAM for large n_bootstraps.
 
         Args:
             X (np.matrix or np.array or pd.Dataframe): a feature matrix
             treatment (np.array or pd.Series): a treatment vector
             y (np.array or pd.Series): an outcome vector
+            p: propensity scores, passed through to fit() if provided
             n_bootstraps (int, optional): number of bootstrap iterations. Default: 200.
             bootstrap_size (int, optional): number of samples per bootstrap. Default: 10000.
             random_state (int, optional): random seed for reproducibility.
             n_jobs (int, optional): number of parallel jobs. -1 uses all cores. Default: 1.
         """
+        from sklearn.base import clone
+
         rng = np.random.RandomState(random_state)
         seeds = rng.randint(0, np.iinfo(np.int32).max, size=n_bootstraps)
         logger.info("Storing bootstrap ensemble ({} iterations)".format(n_bootstraps))
@@ -110,27 +116,15 @@ class BaseLearner(metaclass=ABCMeta):
         def _fit_one(seed):
             local_rng = np.random.RandomState(seed)
             idxs = local_rng.choice(np.arange(X.shape[0]), size=bootstrap_size)
-            X_b, treatment_b, y_b = X[idxs], treatment[idxs], y[idxs]
-            models_c_b = {group: clone(self.model_c) for group in self.t_groups}
-            models_t_b = {group: clone(self.model_t) for group in self.t_groups}
-            for group in self.t_groups:
-                mask = (treatment_b == group) | (treatment_b == self.control_name)
-                treatment_filt = treatment_b[mask]
-                X_filt = X_b[mask]
-                y_filt = y_b[mask]
-                w = (treatment_filt == group).astype(int)
-                if w.sum() == 0 or (w == 0).sum() == 0:
-                    logger.warning(
-                        "Bootstrap sample has no treated or no control units "
-                        "for group {}. Falling back to global model — "
-                        "CI may be underestimated.".format(group)
-                    )
-                    models_c_b[group] = self.models_c[group]
-                    models_t_b[group] = self.models_t[group]
-                    continue
-                models_c_b[group].fit(X_filt[w == 0], y_filt[w == 0])
-                models_t_b[group].fit(X_filt[w == 1], y_filt[w == 1])
-            return models_c_b, models_t_b
+            X_b = X[idxs]
+            treatment_b = treatment[idxs]
+            y_b = y[idxs]
+            p_b = (
+                {group: _p[idxs] for group, _p in p.items()} if p is not None else None
+            )
+            learner_b = clone(self, safe=False)
+            learner_b.fit(X=X_b, treatment=treatment_b, y=y_b, p=p_b)
+            return learner_b
 
         self.bootstrap_models_ = Parallel(n_jobs=n_jobs)(
             delayed(_fit_one)(s) for s in tqdm(seeds)
--- a/causalml/inference/meta/base.py
+++ b/causalml/inference/meta/base.py
@@ -2,6 +2,9 @@ from abc import ABCMeta, abstractmethod
 import logging
 import numpy as np
 import pandas as pd
+from joblib import Parallel, delayed
+from sklearn.base import clone
+from tqdm import tqdm
 
 from causalml.inference.meta.explainer import Explainer
 from causalml.inference.meta.utils import check_p_conditions, convert_pd_to_np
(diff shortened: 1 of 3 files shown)


jeongyoonlee (commented):
Thanks for the PR! The motivation — scoring new unseen data with CIs without re-fitting — is a real use case that the existing `fit_predict()` bootstrap doesn't cover. A few items to address:

## Blocking

1. **`AttributeError` bug** — `self.bootstrap_models_` is never initialized in `__init__`. If a user calls `predict(return_ci=True)` on a freshly constructed (or non-bootstrap-fitted) instance, it raises `AttributeError` instead of the intended `ValueError`. Fix: initialize `self.bootstrap_models_ = None` in `__init__` and use `self.bootstrap_models_ is None` (explicit check) instead of `not self.bootstrap_models_` (falsy check).

2. **No reproducibility** — The bootstrap loop uses `np.random.choice` with no seed. The existing `BaseDRLearner.bootstrap()` already supports a seed pattern — please follow that for consistency.

3. **`ci_quantile` vs `ate_alpha` inconsistency** — The rest of the codebase uses `self.ate_alpha` (set at `__init__` time) to control CI width. This PR introduces a separate `ci_quantile` defaulting to `0.05`, so the same learner produces different CI widths from `fit_predict()` vs `predict(return_ci=True)` unless the user remembers to align them. Recommendation: use `self.ate_alpha` by default, with an optional override parameter.

4. **Duplicated bootstrap logic** — The PR reimplements ~20 lines of resampling/fitting that `BaseLearner.bootstrap()` already handles. Consider reusing or extending the existing infrastructure to reduce maintenance burden.

5. **`return_ci` + `return_components` interaction** — If both are `True`, `return_components` is silently ignored. Should either raise an error or support both.

6. **Flaky test assertion** — `assert (lb <= tau_pred).all() and (tau_pred <= ub).all()` is not guaranteed for bootstrap percentile CIs — the full-sample point estimate can legitimately fall outside the bootstrap percentile interval. Consider checking that a high fraction (e.g., 95%) of predictions fall within bounds, or just check shapes and that `lb <= ub`.

7. **Test conventions** — Please use `RANDOM_SEED` and `CONTROL_NAME` from `tests/const.py` per project conventions, and use `pytest.raises(ValueError)` instead of try/except.

## Non-blocking suggestions

- **Memory cost** — Default 1000 bootstraps storing full deep-copied model pairs can be significant (e.g., 5 treatments × 1000 bootstraps = 10k model objects). Consider documenting this or lowering the default.

- **`BaseTClassifier` not updated** — `BaseTClassifier` overrides `predict()` with its own implementation. Passing `return_ci=True` to a classifier will `TypeError`. Worth documenting or handling.

- **Docstrings** — The new parameters (`store_bootstraps`, `n_bootstraps`, `bootstrap_size`, `return_ci`, `ci_quantile`) should be documented in the `fit()` and `predict()` docstrings.

- **Architecture note** — You mentioned planning to extend this to other learners. Since the store-and-reuse pattern is generic, it might be worth designing the base class integration upfront (even if implementation is phased) to avoid divergent implementations across learners.

aman-coder03:
@jeongyoonlee can you have a look at the PR?
i have addressed all the comments!

aman-coder03:
@jeongyoonlee can you have a look at the PR?
i have addressed all the comments!!!

jeongyoonlee (changes requested):
Thanks for the PR @aman-coder03 — this solves a real pain point (train once, score repeatedly with CIs). A few items to address:

For reference, I cross-checked against `grf` and EconML's approaches:
- **grf** uses [infinitesimal jackknife](https://grf-labs.github.io/grf/reference/predict.causal_forest.html) variance from tree-level predictions — no stored models. Only works for forest-based estimators.
- **EconML** uses [`BootstrapEstimator`](https://github.com/py-why/EconML/blob/main/econml/inference/_bootstrap.py) which stores N cloned models — same design as this PR. Key differences noted below.

---

### 1. Duplicates bootstrap logic — should reuse `BaseLearner.bootstrap()`

The existing `BaseLearner.bootstrap()` in `base.py:55-71` already handles bootstrap sampling, fitting, and predicting on the full population. The new code in `fit()` reimplements this from scratch (sampling, filtering by treatment group, fitting control/treatment models), creating two divergent code paths that must be kept in sync.

**Suggestion**: Reuse `self.bootstrap()` or factor the shared logic into a common method.

### 2. Use `sklearn.base.clone()` instead of `deepcopy`

Both `grf` and EconML use purpose-built clone functions rather than `deepcopy`. `sklearn.base.clone()` creates an unfitted copy with the same hyperparameters and is faster:

```python
from sklearn.base import clone
models_c_b = {group: clone(self.model_c) for group in self.t_groups}
```

### 3. Memory: default `n_bootstraps=1000` is high

With 1000 bootstraps (default), a GBM-based T-learner with 3 treatment groups stores 6000 model objects — easily gigabytes of RAM. EconML's `BootstrapEstimator` has the same trade-off but defaults to 100. Consider lowering the default to 200 and adding a docstring note about memory usage.

### 4. Silent fallback on empty treatment/control groups biases CIs

Lines 133-136: when a bootstrap sample has no treated or no control units for a group, it silently falls back to the global model:
```python
if w.sum() == 0 or (w == 0).sum() == 0:
    models_c_b[group] = self.models_c[group]
    ...
```
This biases CIs toward the point estimate and underestimates variance. Should at minimum log a warning, especially since this can happen frequently with small samples or rare treatments.

### 5. No parallelism

EconML uses `joblib.Parallel` for bootstrap fitting. The sequential loop here will be slow for heavy base learners (e.g., GBM with 1000 bootstraps). Consider adding `n_jobs` support, consistent with the existing `n_jobs` patterns in the meta-learner API.

jeongyoonlee:
Thanks for the quick turnaround. The only feedback I have is the duplicate bootstrap logic between `BaseLearner` and this one. If we update the bootstrap logic only for the T-learner, there will be a discrepancy between the T-learner and other meta-learners. Can you see if we can move your new bootstrap logic to `BaseLearner`?

aman-coder03:
@jeongyoonlee i have moved the bootstrap ensemble logic to `BaseLearner.fit_bootstrap_ensemble()` in `base.py`
`BaseTLearner.fit()` now delegates to it via `self.fit_bootstrap_ensemble(...)`....no duplicate code
this also means the other meta-learners can reuse the same method when we extend the feature to them

causalml/inference/meta/base.py
```diff
@@ -70,6 +73,69 @@ def bootstrap(self, X, treatment, y, p=None, size=10000, rng=None):
...
+                        "Bootstrap sample has no treated or no control units "
+                        "for group {}. Falling back to global model — "
+                        "CI may be underestimated.".format(group)
+                    )
+                    models_c_b[group] = self.models_c[group]
+                    models_t_b[group] = self.models_t[group]
+                    continue
+                models_c_b[group].fit(X_filt[w == 0], y_filt[w == 0])
+                models_t_b[group].fit(X_filt[w == 1], y_filt[w == 1])
+            return models_c_b, models_t_b
```
jeongyoonlee: This routine is T-learner specific. The bootstrap logic in BaseLearner should be applicable across all meta-learners. Check EconML's [BootstrapEstimator](https://www.pywhy.org/EconML/_modules/econml/inference/_bootstrap.html#BootstrapEstimator) for a reference.

jeongyoonlee (changes requested):

jeongyoonlee:
Thanks for the iteration @aman-coder03 — the design is in good shape now. A few items still to address before merge:

### 1. Closure passed to `joblib.Parallel` will fail with `n_jobs != 1`

`_fit_one` in `fit_bootstrap_ensemble` (`base.py:113-126`) is a nested function closing over `X`, `treatment`, `y`, `p`. With the default `n_jobs=1` joblib runs serially and this works, but with `n_jobs > 1` it switches to the loky backend and pickling closures is unreliable across cloudpickle versions. This is a known gotcha in this codebase — see CLAUDE.md.

Please refactor `_fit_one` to a module-level function (e.g. `_fit_bootstrap_clone(learner_template, X, treatment, y, p, seed, bootstrap_size)`) and pass the data as arguments via `delayed(...)`.

### 2. Duplicate `clone` import

`from sklearn.base import clone` is imported at the top of `base.py` (line 6) and again inside `fit_bootstrap_ensemble` (line 105). Drop the local import.

### 3. Test coverage gaps

The reproducibility and shape tests are good, but please add:
- A test with `n_jobs=2` (or `-1`) to lock in parallel execution and catch the closure issue above.
- An assertion that two runs with *different* `random_state` values produce *different* bounds (catches accidentally constant seeds).
- A test that exercises `BaseTClassifier` with `return_ci=True` (will currently `TypeError` — see the separate issue).

I'll file a separate issue for two non-blocking follow-ups so they don't expand the scope of this PR:
- `BaseTClassifier` raises `TypeError` instead of a meaningful error on `return_ci=True`.
- `clone(self, safe=False)` in `fit_bootstrap_ensemble` silently falls back to `deepcopy` of the fitted learner, which is misleading vs. the EconML-pattern claim in the docstring.

jeongyoonlee:
Re-reviewed against `a31b8d3` — most prior items are addressed (closure → module-level helper, lint green, parallelism + reproducibility tests). One blocker remains:

**`BaseTClassifier` will `TypeError` on `predict(return_ci=True)`.** It inherits the new `fit(store_bootstraps=True)` from `BaseTLearner` but overrides `predict()` (`tlearner.py:441`) without a `return_ci` parameter, so:

```python
learner = BaseTClassifier(learner=LogisticRegression())
learner.fit(X, treatment, y, store_bootstraps=True)
learner.predict(X, return_ci=True)   # TypeError: unexpected keyword 'return_ci'
```

The percentile logic is identical — the bootstrap loop just calls `learner_b.predict(X)` and the cloned classifier's `predict()` returns the same `[n_samples, n_treatment]` `te` matrix — so mirroring the `if return_ci:` block in `BaseTClassifier.predict()` should be a small change. Alternatively, guard at `fit()` time with a clear error if you want to keep the classifier work out of scope.

aman-coder03:
@jeongyoonlee i have added `return_ci` support to `BaseTClassifier.predict()` in the latest commit, mirrors the same logic as `BaseTLearner.` Please take a look!

jeongyoonlee (changes requested):
Thanks for the iteration @aman-coder03 — two blockers in the latest commit (`ee7d69e`):

### 1. Duplicated dead-code block in `BaseTLearner.predict()`
`tlearner.py:198-228` — the `if return_ci and return_components:` check and the `if return_ci:` CI block appear **twice consecutively**. The second copy is unreachable (first block returns), but it's a confusing rebase artifact:

```python
if return_ci:                       # block A
    ...
    return te, te_lower, te_upper   # exits here
if return_ci and return_components: # dead
    raise ValueError(...)
if return_ci:                       # dead — identical to block A
    ...
    return te, te_lower, te_upper
```
Delete the second copy. The early-return guard at the top of the function already handles the `return_ci + return_components` case.

### 2. `BaseTClassifier.predict()` silently ignores `return_ci=True`
The latest commit added `return_ci=False` to the signature at `tlearner.py:456-466`, but the function body never branches on it. So:
```python
clf.fit(X, t, y, store_bootstraps=True)
tau, lb, ub = clf.predict(X, return_ci=True)   # ValueError on unpack
```
returns just `te` instead of `(te, te_lower, te_upper)` — silent contract violation, worse than the previous `TypeError`.

Two reasonable fixes:
- (Preferred) Lift the CI block in `BaseTLearner.predict()` into a small helper (e.g. `_compute_bootstrap_ci(X)`) and call it from both predict paths — also resolves (1).
- Otherwise, `raise NotImplementedError("return_ci not yet supported for BaseTClassifier")` so failure is loud.

Add a `BaseTClassifier` variant of the new test using `generate_classification_data` either way — that gap is what let the silent-ignore through.

CI is green otherwise.

jeongyoonlee (approved):
Both blockers fixed cleanly in `de0e3602`. LGTM — approving.

**One request before merge:** add a `BaseTClassifier` variant of `test_BaseTLearner_predict_return_ci`. The classifier path is still untested, which is what let the silent-ignore through last round.

Two small follow-ups (predict() validation ordering; `clone(safe=False)` deepcopy fallback) filed separately so they don't gate this PR.

PR #890, opened 2026-04-05
Add `rate_score()` with bootstrap confidence intervals and p-values

### LINKED ISSUE #889: Add bootstrap confidence intervals and p-values to `rate_score()`
current `rate_score()` returns point estimates only. Yadlowsky et al. (2021) derive a functional CLT enabling exact Gaussian inference via half-sample bootstrap, which is also implemented in the grf R package via `rank_average_treatment_effect()`

proposed extension is to add a `return_ci=True` parameter to `rate_score()`...
```
    rate_score(df, weighting="autoc", return_ci=True, n_bootstrap=200)
```
which would return the RATE estimate along with standard errors and p-values.

related to #887

aman-coder03:
## Proposed changes

closes #889
extends the existing `rate_score()` in `causalml/metrics/rate.py` (added in #887) with bootstrap confidence intervals and p-values, as requested by @jeongyoonlee in the #887 review.

### What's added

- new parameters: `return_ci=False`, `n_bootstrap=200`, `alpha=0.05`
- when `return_ci=True`, uses half-sample bootstrap (m = n // 2, without replacement) to estimate standard errors and confidence intervals
- p-value tests H0: RATE = 0 via two-sided z-test
- refactored the integration logic into a `_compute_rate_from_toc()` helper to avoid code duplication between the point estimate and bootstrap paths
- 4 new tests added to `tests/test_rate.py`

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/tests/test_rate.py
+++ b/tests/test_rate.py
@@ -161,65 +161,50 @@ def test_plot_toc(synthetic_df, monkeypatch):
     assert isinstance(ax, plt.Axes)
 
 
-def test_rate_score_return_ci_returns_dataframe():
-    np.random.seed(42)
-    n = 300
-    df = pd.DataFrame(
-        {
-            "y": np.random.binomial(1, 0.5, n),
-            "w": np.random.binomial(1, 0.5, n),
-            "tau": np.random.normal(0.5, 1, n),
-            "model": np.random.normal(1, 1, n),
-        }
+def test_rate_score_return_ci_returns_dataframe(synthetic_df):
+    result = rate_score(
+        synthetic_df,
+        treatment_effect_col="tau",
+        return_ci=True,
+        n_bootstrap=50,
+        random_state=RANDOM_SEED,
     )
-    result = rate_score(df, return_ci=True, n_bootstrap=50)
     assert isinstance(result, pd.DataFrame)
     assert set(result.columns) == {"rate", "se", "ci_lower", "ci_upper", "p_value"}
 
 
-def test_rate_score_ci_bounds_ordered():
-    np.random.seed(42)
-    n = 300
-    df = pd.DataFrame(
-        {
-            "y": np.random.binomial(1, 0.5, n),
-            "w": np.random.binomial(1, 0.5, n),
-            "tau": np.random.normal(0.5, 1, n),
-            "model": np.random.normal(1, 1, n),
-        }
+def test_rate_score_ci_bounds_ordered(synthetic_df):
+    result = rate_score(
+        synthetic_df,
+        treatment_effect_col="tau",
+        return_ci=True,
+        n_bootstrap=50,
+        random_state=RANDOM_SEED,
     )
-    result = rate_score(df, return_ci=True, n_bootstrap=50)
     assert (result["ci_lower"] < result["rate"]).all()
     assert (result["ci_upper"] > result["rate"]).all()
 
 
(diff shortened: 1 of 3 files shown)


aman-coder03:
@jeongyoonlee can you have a look at the PR??

jeongyoonlee (changes requested):
Thanks for the PR @aman-coder03 — the half-sample bootstrap approach is solid and aligns with Yadlowsky et al. (2021). A few items to address before merge:

---

### 1. Hardcoded random seed

`rng = np.random.default_rng(seed=42)` is hardcoded. Users can't vary bootstrap randomness or reproduce results with a different seed.

**Fix**: Add a `random_state=None` parameter (consistent with sklearn conventions) and pass it to `default_rng()`.

### 2. `_compute_rate_from_toc` should be module-level

The helper is defined as a nested function inside `rate_score()`. It gets re-created on every call and closures can cause pickle issues if this function is ever used with joblib. Move it to module level (prefixed with `_`).

### 3. `from scipy import stats` should be a top-level import

scipy is already a required dependency. Move the import to the top of the file for consistency with the rest of the codebase.

### 4. Tests should use existing fixtures and `RANDOM_SEED`

The 4 new tests each create their own DataFrame with `np.random.seed(42)` instead of:
- Using the `synthetic_df` or `rct_df` fixtures already defined in the file
- Using `RANDOM_SEED` from `tests/const.py` and `np.random.default_rng()`

Please refactor to use the existing fixtures where possible.

### 5. Missing backward-compatibility test

The refactoring into `_compute_rate_from_toc()` should be verified to produce numerically identical results to the old inline code. Add a test that `return_ci=False` (default) returns a `pd.Series` matching the pre-PR behavior.

### 6. Missing test for observed-outcome mode with CI

All 4 new tests pass `tau` (oracle mode). None test the observed-outcome path (`outcome_col`/`treatment_col` without `treatment_effect_col`), which is the primary use case. Add at least one test using the `rct_df` fixture.

### 7. Lint error

Please fix the lint error with `black .`

---

Overall the approach is correct and well-documented. The main asks are: expose `random_state`, use existing test fixtures, and add coverage for the observed-outcome path.

tests/test_rate.py
```diff
@@ -159,3 +159,67 @@ def test_plot_toc(synthetic_df, monkeypatch):
     monkeypatch.setattr(plt, "show", lambda: None)
     ax = plot_toc(synthetic_df, treatment_effect_col="tau")
     assert isinstance(ax, plt.Axes)
+
+
+def test_rate_score_return_ci_returns_dataframe():
+    np.random.seed(42)
```
jeongyoonlee: Can you use RANDOM_SEED instead?

tests/test_rate.py
```diff
@@ -159,3 +159,67 @@ def test_plot_toc(synthetic_df, monkeypatch):
...
+    n = 300
+    df = pd.DataFrame(
```
jeongyoonlee: Let's use the `synthetic_df` or `rct_df` fixture defined above.

jeongyoonlee (changes requested):
Thanks for the quick turnaround. Can you address the minor comments in the test file and rebase the code onto the latest master to resolve the merge conflict? Otherwise, it looks good.

jeongyoonlee (approved):
LGTM. Thanks!

PR #901, opened 2026-05-23
Add native Polars DataFrame, Series, and LazyFrame support for all meta-learners

### LINKED ISSUE #855: Polars support on CausalML
Add polars support to CausalML

**Additional context**
CausalML supports pandas, but doesn't support polars

aman-coder03:
## Proposed changes
Closes #855

Adds native Polars support to CausalML, keeping feature matrices (`X`) in their native format (numpy / pandas / polars) end-to-end across all five meta-learners (T, S, X, R, DR)

## What changed
- `causalml/inference/meta/utils.py` added `collect_if_lazy(X)`, `n_rows(X)`, `filter_mask()`, `filter_index()`, `prepend_column()`, `concat_treatment_col()`, and `to_numpy()`. Kept `convert_pd_to_np` as a deprecated backward-compat alias for `explainer.py` and other existing callers.
- `causalml/inference/meta/base.py`.... `bootstrap()` resamples via `filter_index(X, idxs)` so X stays native; `_format_p()` uses `to_numpy`; `_set_propensity_models()` filters X natively and passes it as-is to sklearn (≥ 1.6 accepts pandas/Polars).
- `causalml/inference/meta/tlearner.py` ....`collect_if_lazy(X)` at the top of every public method; `model_c` fitted once on the full control set and exposed as a shared-reference dict; `store_bootstraps`/`return_ci` API merged from upstream #904.
- `causalml/inference/meta/slearner.py`....`np.hstack` replaced with `concat_treatment_col`/`prepend_column` (type-safe across numpy/pandas/polars).
- `causalml/inference/meta/xlearner.py` .... `model_mu_c` fitted once on the full control set (regressor: `.predict`; classifier: `.predict_proba`); exposed as a shared-reference dict; `self.var_c` stored as a finite scalar.
- `causalml/inference/meta/rlearner.py`.... `cross_val_predict` receives native X; `treatment`/`y`/`sample_weight` normalised to numpy at entry.
- `causalml/inference/meta/drlearner.py` .... KFold partition slices use `filter_index(X, idx)` so X stays native across all three cross-fit folds.
- `causalml/propensity.py`.... removed top-level import of `convert_pd_to_np` (caused circular import on cold import of `causalml.propensity`); `PropensityModel.fit/predict` and `compute_propensity_score` pass X straight through to sklearn/XGBoost which accept pandas/Polars natively.

## Design decisions
- **Polars is an optional dependency** ....`try/except ImportError` guard in `utils.py`; `polars>=1.0.0` declared as `[project.optional-dependencies] polars` and pulled into `[test]` so CI runs the suite
- **`pl.LazyFrame` support** ....collected once to `pl.DataFrame` at the public-method boundary via `collect_if_lazy()`; all helpers work on `pl.DataFrame`/`pl.Series` only
- **Return types unchanged** .... all methods still return numpy arrays
- **LightGBM caveat** .... LightGBM has a known sklearn-API bug with Polars (lightgbm-org/LightGBM#6849); users should convert X to numpy/pandas when using LightGBM as a base learner

## Testing
Added `tests/test_polars_support.py` with coverage across all five learner types and both regressors and classifiers (BaseTClassifier, BaseSClassifier, BaseXClassifier), verifying that numpy/pandas/polars/LazyFrame inputs produce identical results, and covering edge cases like mixed inputs and fit-on-numpy/predict-on-polars.

## Types of changes

## Checklist

## Further comments


jeongyoonlee:
Thanks for looking into this, @aman-coder03. 

As I commented on #855, converting the dataframe into numpy is not ideal for Polars, as it doesn't benefit from its performant features. Instead, adding support for native Pandas and Polars DataFrames is recommended. It will require careful inspection of the current indexing to make it comparable to the indexing on DataFrames.

aman-coder03:
thanks for the feedback @jeongyoonlee 

i understand the concern....converting to NumPy at the boundary means Polars users lose all the performance benefits(lazy evaluation, zero-copy operations, columnar efficiency) that they came for in the first place

for the native approach, my plan would be...
- add thin abstraction layer in `utils.py` with helpers like `filter_by_mask()`, `concat_cols()`, and `get_values()` that dispatch to the correct pandas/polars/numpy operation based on the input type
- audit every indexing pattern across all five learner files(`tlearner.py`, `slearner.py`, `xlearner.py`, `rlearner.py`, `drlearner.py`) then replace them with these helpers
- key patterns to handle are: boolean mask filtering (`X[mask]`), column concatenation (`np.hstack`), and in-place mutation(Polars is immutable)
- keep `convert_pd_to_np()` only at the final NumPy-only boundaries(sklearn's `cross_val_predict` in the R-learner), since `sklearn` doesn't accept Polars natively

Before I go ahead and rewrite, a couple of questions to make sure i am heading in the right direction...
1. should the output of `predict()` remain a NumPy array, or would you like an option to return a Polars DataFrame when the input was Polars?
2. for `sklearn` calls like `cross_val_predict` (R-learner) and propensity model fitting, converting to NumPy at that specific boundary seems unavoidable....does that approach work for you, or do you have a preferred alternative?

Happy to update the PR once we are aligned

jeongyoonlee:
Thanks for the detailed plan, @aman-coder03. One important correction before you start: **converting to NumPy at the sklearn boundary isn't necessary.**

- **scikit-learn ≥ 1.4** accepts Polars (and pandas) DataFrames natively via the DataFrame Interchange Protocol ([release notes](https://scikit-learn.org/stable/whats_new/v1.4.html)); causalml pins `scikit-learn>=1.6.0`.
- **XGBoost ≥ 3.1** accepts `pl.DataFrame` and `pl.LazyFrame` directly ([docs](https://xgboost.readthedocs.io/en/stable/python/python_intro.html)).
- **LightGBM** nominally accepts `pl.DataFrame` but has a known sklearn-API bug ([issue #6849](https://github.com/microsoft/LightGBM/issues/6849)) — document as a caveat.

So the rewrite scope collapses to causalml's own indexing/concat/mutation — not the model calls.

Answers to your questions:

1. **`predict()` output type.** Keep returning NumPy. The entire underlying stack does the same: scikit-learn's `.predict()` returns NumPy regardless of input (only transformers honor `set_output("polars")`, not predictors), XGBoost's `.predict()` returns NumPy (or cupy.ndarray on GPU) even when fed a `pl.DataFrame` via its zero-copy Arrow path, and LightGBM's `.predict()` returns NumPy as documented. Matching that convention keeps causalml's API consistent with sklearn/xgboost/lightgbm and preserves backward compatibility. A `return_type="polars"` opt-in can be added later if users ask.

2. **NumPy conversion at sklearn boundaries.** Skip it — pass Polars/pandas straight through. Zero materialization overhead, and dtype/feature-name plumbing is handled by the ML library.

Suggested phasing:

- **Phase 1**: Remove `convert_pd_to_np()` in favor of native pandas DataFrame support across all meta-learners. No Polars yet — this isolates the indexing-polymorphism work and gives a clean baseline.
- **Phase 2**: Add Polars support to one estimator (e.g. `BaseTLearner`) as a reference implementation with equivalence tests (NumPy / pandas / Polars produce identical `te`).
- **Phase 3+**: Extend Polars to the remaining learners (S/X/R/DR), one PR each.

A few practical notes for the audit:

- **Enumerate the actual patterns first.** Grep all five learners for indexing/concat/mutation (`X[mask]`, `np.hstack`, `.iloc`, in-place assignment) and design the minimal helper set from real usage, not guesses.
- **Polars is immutable** — anywhere current code does `X[mask] = value` or accumulates by index assignment, you'll need an explicit rewrite (`with_columns`, `when/then/otherwise`), not just a dispatch helper.
- **Add `polars` to test deps in Phase 2.** The current PR's tests don't run in CI because Polars isn't installed there — `pytest.importorskip("polars")` skips the whole file silently.

Happy to review Phase 1 once it's ready.

aman-coder03:
thanks for the detailed feedback @jeongyoonlee 
I understand the phasing suggestion, but I've intentionally tried to complete the full Polars support in this single PR....covering all five learners (T/S/X/R/DR) together. My reasoning was that the indexing patterns are largely the same across learners, so the helper set (filter_mask, filter_index, prepend_column, etc.) naturally fell out of auditing all of them at once rather than discovering gaps PR by PR.

i'm happy to reorganize if you'd prefer the phased approach...I can split this into separate PRs for Phase 1 (pandas cleanup) and Phase 2/3 (Polars per learner). But if the current implementation looks sound to you, I'd love to land it as is to avoid rebasing overhead

jeongyoonlee (changes requested):
Thanks for pushing this through to a full implementation, @aman-coder03 — the helper-based
dispatch (`filter_mask` / `filter_index` / `prepend_column` / `concat_treatment_col`) is the
right abstraction, and the T-learner is a good reference (docstrings kept + updated, explicit
`LazyFrame.collect()` guard). The direction is solid.

One process note before the specifics: this landed all of Phase 1–3 plus a large cosmetic
refactor in a single ~1k-line PR, which makes it hard to review and bisect. I'm not going to
ask you to re-split it now, but the four items below are blocking regardless of how it's
packaged.

## Blocking

### 1. Circular import — `import causalml.propensity` crashes on a cold import
`propensity.py` adds a top-level `from causalml.inference.meta.utils import convert_pd_to_np`.
Importing that submodule runs `meta/__init__.py` → `slearner` → `base.py` (`from
causalml.propensity import compute_propensity_score`) back into the half-initialized
`propensity` module. Reproduced by applying just that one line:

```
File ".../causalml/propensity.py", line 11, in <module>
    from causalml.inference.meta.utils import convert_pd_to_np
File ".../causalml/inference/meta/__init__.py", line 1, in <module>
    from .slearner import LRSRegressor, BaseSLearner, BaseSRegressor, BaseSClassifier
File ".../causalml/inference/meta/slearner.py", line 9, in <module>
    from causalml.inference.meta.base import BaseLearner
File ".../causalml/inference/meta/base.py", line 8, in <module>
    from causalml.propensity import compute_propensity_score
ImportError: cannot import name 'compute_propensity_score' from partially initialized
module 'causalml.propensity' (most likely due to a circular import)
```

Both `import causalml.propensity` and `from causalml.propensity import
ElasticNetPropensityModel` fail cold. It passes locally only because the test process imports
`causalml.inference.meta` first, which caches it before `propensity` runs its new import — so
the cycle never re-triggers under pytest, but it breaks normal user entry points.

Fix: make the import function-local in the three methods that use it, or drop it — sklearn
≥1.6 accepts pandas/Polars natively, so `PropensityModel.fit`/`predict` may not need the
conversion at all.

### 2. `polars` is not declared as a dependency → the test file is skipped in CI
`pyproject.toml` is untouched, and `tests/test_polars_support.py` starts with
`pytest.importorskip("polars")`, so the entire feature is silently skipped in CI (same point
from the 2026-05-28 round). The code is already optional-ready (try/except in `utils.py`,
guarded `import polars` in `tlearner.py`), so keep polars **optional** — don't add it to core
`[project.dependencies]`. Add an optional extra and wire it into `test` so CI (which installs
`-e ".[test]"`) actually runs the suite:

```toml
[project.optional-dependencies]
polars = ["polars>=1.0.0"]      # users: pip install causalml[polars]
test = [
    "pytest>=4.6",
    "pytest-cov>=4.0",
    "causalml[polars]",          # self-referencing extra so .[test] pulls polars in CI
]
```

A dev without polars still gets a clean `importorskip` skip; no `--runpolars` flag needed.

### 3. Docstrings deleted across S/X/R/DR
Every class / `__init__` / `fit` / `predict` / `fit_predict` / `estimate_ate` docstring in
`slearner.py`, `xlearner.py`, `rlearner.py`, `drlearner.py` was removed, including the paper
references (Kennedy 2020, Nie & Wager 2019, Künzel et al. 2018). These render on readthedocs.
The T-learner correctly *kept and updated* its docstrings to mention `pl.DataFrame` — please
follow that pattern for the other four rather than deleting.

### 4. LazyFrame handling is inconsistent — and it points at the core design rule
Only the T-learner guards `predict` with `if isinstance(X, pl.LazyFrame): X = X.collect()`.
In S/X/R/DR, `predict` passes X to `prepend_column` / `model.predict` directly — and
`prepend_column`/`concat_treatment_col` have no `LazyFrame` branch and call `len(X)`, which
raises on a `pl.LazyFrame`. It's untested because the only `test_lazyframe_input` lives in
`TestTLearnerPolars`.

We want to **keep native end-to-end** (that's the whole point — down-converting X to numpy
would throw away the Polars benefit, and matters in particular for the xgboost zero-copy path
and column-name-aware `Pipeline`/`ColumnTransformer` learners). The fix is to apply the
contract *consistently*:

> **X stays native end-to-end; `treatment` / `y` / `p` / `sample_weight` normalize to numpy at
> entry.** Those are 1-D vectors that masking/`np.unique`/`.astype` need, and they're unrelated
> to the wide-frame promise.

Concretely:
- **Collect `LazyFrame` once at the top of each public method** into a `pl.DataFrame` (not
  numpy). You have to collect to row-mask anyway, so this is the natural single point. After
  that, the helpers only need to handle `pl.DataFrame`/`pl.Series`, and S/X/R/DR get LazyFrame
  support for free.
- **Never `to_numpy(X)` just to read a row count** — `drlearner.bootstrap` does
  `to_numpy(X).shape[0]`, and the `te = np.zeros((X.shape[0], ...))` allocations should read
  the count natively (`X.shape[0]` / `len(X)` work for numpy/pandas/polars).
- Let `bootstrap` resample on the native X via `filter_index` (callers currently pre-convert
  to `X_np` before calling it).
- Tests to lock the contract in: numpy == pandas == polars equivalence for **every** learner
  (regressor *and* classifier); a fake learner asserting `isinstance(X, pl.DataFrame)` inside
  `fit`/`predict` so a reintroduced `to_numpy(X)` fails loudly; a no-feature-name-warning
  assertion on the fit-DataFrame/predict-DataFrame path; and a by-name `Pipeline` learner.

Note: lightgbm still has the sklearn-API Polars bug (microsoft/LightGBM#6849), so native will
break it — document as a caveat or convert only at the lightgbm boundary.

---

Non-blocking polish (smaller correctness + test-coverage items) is in a follow-up note so it
doesn't clutter the merge gate. Happy to re-review quickly once the four above are addressed.

jeongyoonlee:
@aman-coder03, can you check and fix the [build error](https://github.com/uber/causalml/actions/runs/27777924342/job/82400602284?pr=901)?

aman-coder03:
done @jeongyoonlee

causalml/inference/meta/slearner.py
```diff
@@ -114,14 +134,23 @@ def predict(
 
         for group in self.t_groups:
             model = self.models[group]
+
+            # Build separate frames for control and treatment to avoid in-place
+            # mutation, which fails when learners like CatBoost set the
+            # writeable flag to False on arrays passed to predict().
+            X_new_c = prepend_column(0.0, X)
```
jeongyoonlee, line 138: Lines 132-133 (`X_new_c`/`X_new_t = np.hstack(...)`) are now dead — they're overwritten by `prepend_column()` here before use — and they force a full numpy copy of X on every `predict()`, which defeats the native-X path this PR adds. Remove them.

causalml/inference/meta/slearner.py
```diff
@@ -114,14 +134,23 @@ def predict(
...
+            yhat_cs[group] = model.predict(X_new_c)
+
+            X_new_t = prepend_column(1.0, X)
             yhat_cs[group] = model.predict(X_new_c)
```
jeongyoonlee, line 140: Duplicate: `yhat_cs[group]` was already computed at line 142, so this re-runs `model.predict(X_new_c)` for nothing. Drop this line.

causalml/inference/meta/slearner.py
```diff
@@ -349,14 +387,20 @@ def predict(
 
         for group in self.t_groups:
             model = self.models[group]
+
+            X_new_c = prepend_column(0.0, X)
+            yhat_cs[group] = model.predict_proba(X_new_c)[:, 1]
+
+            X_new_t = prepend_column(1.0, X)
             yhat_cs[group] = model.predict_proba(X_new_c)[:, 1]
```
jeongyoonlee, line 387: Same duplicate as the regressor: `yhat_cs[group]` was set at line 392; this repeats `predict_proba(X_new_c)`. Drop it.

causalml/inference/meta/drlearner.py
```diff
@@ -225,15 +259,18 @@ def predict(
...
             verbose (bool, optional): whether to output progress logs
         Returns:
             (numpy.ndarray): Predictions of treatment effects.
         """
-        X, treatment, y = convert_pd_to_np(X, treatment, y)
+        X = collect_if_lazy(X)
 
+        te = np.zeros((n_rows(X), self.t_groups.shape[0]))
+        yhat_cs = {}
         te = np.zeros((X.shape[0], self.t_groups.shape[0]))
```
jeongyoonlee: Duplicate `te` allocation: line 272 already set it via the Polars-safe `n_rows(X)`, and this line overwrites it with `X.shape[0]`. Remove this line and keep 272.

causalml/inference/meta/drlearner.py
```diff
@@ -593,8 +629,10 @@ def predict(
                 - dict: Predicted probabilities for the control group (yhat_cs).
                 - dict: Predicted probabilities for the treatment group (yhat_ts).
         """
-        X, treatment, y = convert_pd_to_np(X, treatment, y)
+        X = collect_if_lazy(X)
 
+        te = np.zeros((n_rows(X), self.t_groups.shape[0]))
+        yhat_cs = {}
         te = np.zeros((X.shape[0], self.t_groups.shape[0]))
```
jeongyoonlee: Same duplicate `te` as `BaseDRLearner.predict`: line 634 already set it via `n_rows(X)`. Remove this line.

causalml/inference/meta/tlearner.py
```diff
@@ -490,19 +507,22 @@ def predict(
...
+                treatment_filt_np = treatment_np[mask]
+                y_filt = to_numpy(filter_mask(y, mask))
+                w = (treatment_filt_np == group).astype(int)
 
                 yhat = np.zeros_like(y_filt, dtype=float)
                 yhat[w == 0] = yhat_c[mask][w == 0]
                 yhat[w == 1] = yhat_ts[group][mask][w == 1]
 
                 logger.info("Error metrics for group {}".format(group))
+                from causalml.metrics import classification_metrics
```
jeongyoonlee: Redundant local import — `classification_metrics` is already imported at module top (line 19). Remove.

causalml/inference/meta/xlearner.py
```diff
@@ -523,25 +528,29 @@ def fit(self, X, treatment, y, p=None):
...
 
         if p is None:
-            self._set_propensity_models(X=X, treatment=treatment, y=y)
+            self._set_propensity_models(X=X, treatment=treatment_np, y=to_numpy(y))
             p = self.propensity
         else:
             p = self._format_p(p, self.t_groups)
 
         self._classes = {group: i for i, group in enumerate(self.t_groups)}
+        self.models_mu_c = {group: deepcopy(self.model_mu_c) for group in self.t_groups}
```
jeongyoonlee: This fits a per-group control model, but every group's mask selects the full control set, so for multi-treatment data you now train N identical control models (the regressor still shares one — line 137). Results are unchanged but it's a perf regression; consider keeping the shared model or noting why it was dropped.

jeongyoonlee (commented):
The native-X approach looks right and matches the existing sklearn >=1.6 floor, but the diff carries some merge cruft and is broader than the feature itself. Two things before merge:

- The PR description is stale: it says the individual learner files weren't changed (all five were), and that `convert_pd_to_np()` calls were added to `propensity.py` fit/predict/compute_propensity_score (the diff adds none — only docstrings and a `.copy()` guard). Please update it to match what shipped.
- The classifier Polars paths (`BaseXClassifier`/`BaseTClassifier`/`BaseSClassifier`) have no tests, yet they're the most heavily rewritten code in this PR. Please add coverage mirroring the regressor tests.

Inline notes below — none are correctness-breaking, but the duplicated/dead lines should be cleaned up.

causalml/inference/meta/xlearner.py
```diff
@@ -51,28 +54,26 @@ def __init__(
             and (treatment_effect_learner is not None)
         )
 
-        if control_outcome_learner is None:
-            self.model_mu_c = deepcopy(learner)
-        else:
-            self.model_mu_c = control_outcome_learner
-
-        # Preserve the unfitted template so repeated fit() calls always start fresh.
-        self._model_mu_c_template = self.model_mu_c
```
jeongyoonlee: We need this for fit() to start fresh.

causalml/inference/meta/xlearner.py
```diff
@@ -124,99 +128,102 @@ def fit(self, X, treatment, y, p=None):
...
-        # group. Deepcopy from the unfitted template so re-calling fit() starts fresh.
-        control_mask = treatment == self.control_name
-        self.model_mu_c = deepcopy(self._model_mu_c_template)
-        self.model_mu_c.fit(X[control_mask], y[control_mask])
-        # Expose as a shared-reference dict to preserve the public models_mu_c API.
+        # model_mu_c is trained on control only (identical across groups) — fit once.
+        control_mask = treatment_np == self.control_name
+        X_control = filter_mask(X, control_mask)
+        y_control = to_numpy(filter_mask(y, control_mask))
+        self.model_mu_c = deepcopy(self.model_mu_c)
```
jeongyoonlee: We should keep master, i.e., deepcopying the unfit model_mu_c_template instead of model_mu_c to start fresh at each call to fit().

jeongyoonlee (changes requested):

causalml/inference/meta/xlearner.py
```diff
@@ -549,103 +557,109 @@ def fit(self, X, treatment, y, p=None):
...
-        # model_mu_c is trained on control data, which is the same for every treatment
-        # group, so fit it once and store as a single model (not a per-group dict).
-        control_mask = treatment == self.control_name
-        self.model_mu_c = deepcopy(self._model_mu_c_template)
-        self.model_mu_c.fit(X[control_mask], y[control_mask])
+        # model_mu_c is trained on control only (identical across groups) — fit once.
+        control_mask = treatment_np == self.control_name
+        X_control = filter_mask(X, control_mask)
+        y_control = filter_mask(y, control_mask)
+        self.model_mu_c = deepcopy(self.model_mu_c)
```
jeongyoonlee: same as above. copy model_mu_c_template instead of model_mu_c

jeongyoonlee (changes requested):
Thanks for the update — re-reviewed at `8a93537`.

**Resolved since the last round:** the `propensity.py` circular import (blocker 1) and the polars dependency declaration (blocker 2 — now an optional extra + self-referenced in the `test` extra, so `importorskip` actually runs in CI).

The latest push, though, looks like it merged the polars rewrite on top of the verbatim-`__init__` refactor without resolving the overlaps, so the branch now carries unresolved merge artifacts — including a crash in the X-learner. These need cleanup before merge.

### Blockers

**1. X-learner `predict`/`estimate_ate` crashes on native Polars (and non-default-index pandas) X.** `xlearner.py:255` (regressor) and `:704` (classifier) keep X native and then index it with a boolean mask:

```python
control_mask = treatment == self.control_name
yhat_c_verbose = self.model_mu_c.predict(X[control_mask])
```

On polars 1.42, `X[bool_mask]` raises (`ValueError: expected N values when selecting columns by boolean mask` for a numpy mask; `TypeError` for a pandas Series mask). This is the default path — `predict(X, treatment, y)` with `verbose=True` — so it defeats the feature for the X-learner. The tests miss it because every X-learner test calls `predict(X)` with no `treatment`/`y`, skipping the branch. `yhat_c_verbose` is also never read; the per-group loop below already computes the verbose metrics via `to_numpy(treatment)` + `filter_mask`. Please delete the preamble (or mirror the loop) and add a test that calls `predict(X_polars, treatment, y)`.

**2. The branch has unresolved merge conflicts — duplicated blocks and code stranded inside docstrings.** Concrete spots:

- `xlearner.py:169-187` — `BaseXLearner.fit` trains the control outcome model twice (a "fit once" block immediately overwritten by an old re-deepcopy+refit block).
- `drlearner.py:108` — `X = collect_if_lazy(X)` sits inside the `fit` docstring, so it never runs.
- `drlearner.py:61-81` (and the X-learner `__init__`) — an eager `self.model_mu_c = deepcopy(...)` block + a dangling `"""…stored verbatim…"""` no-op string before the actual verbatim stores; the eager attrs are dead.
- `rlearner.py:658-666` — `XGBRRegressor.fit` calls `check_treatment_vector` twice and has a dangling `"""Fit using early-stopping…"""` string.
- `rlearner.py:532-541` — `BaseRClassifier.fit` computes `vars_c`/`vars_t` twice.
- `drlearner.py:641`, `xlearner.py:685` — dangling docstring statements in the classifier `predict` overrides.

A clean re-merge (polars rewrite vs. `__init__` refactor) should remove all of these.

**3. X/R/DR don't actually keep X native (re: prior blocker 4).** These three still call `convert_pd_to_np(X, …)` in `fit`, converting X to numpy, while the new docstrings claim "the feature matrix is otherwise kept in its native format throughout." Either keep X native end-to-end (as S/T do) or correct the docstrings so the contract matches the code.

A few non-blocking polish items I'll keep out of the thread for now. Happy to walk through any of the above.

jeongyoonlee:
Thanks for the "fixing blockers" push, @aman-coder03 — re-reviewed at `509a213`. Most of the previous round is resolved: the X-learner verbose-predict crash on native `X`, the `model_mu_c` double-fit, the deleted S/X/R/DR docstrings (paper refs included), and the leftover merge artifacts are all cleared. Nice.

One blocker remains, plus a few non-blocking nits.

### Blocker — `_fit_bootstrap_clone` still crashes on a native (pandas/polars) `X`

`bootstrap()` was correctly switched to `n_rows(X)` + `filter_index(X, idxs)`, but its module-level twin `_fit_bootstrap_clone` (the `store_bootstraps=True` → `predict(return_ci=True)` path) was missed:

```python
# causalml/inference/meta/base.py:39-40
idxs = rng.choice(np.arange(X.shape[0]), size=bootstrap_size)
X_b = X[idxs]
```

Since the keep-native refactor dropped the entry-point conversion, `tlearner.fit(..., store_bootstraps=True)` now forwards a native `X` straight here (`tlearner.py:144` passes `X=X`; only `treatment`/`y` are converted). `X[idxs]` then breaks — and this is a regression for **pandas** users too, not only polars:

```
>>> pd.DataFrame({"a":[1,2,3]})[np.array([0,2,1])]
KeyError: "None of [Index([0, 2, 1], dtype='int64')] are in the [columns]"
```

Fix: route it through `n_rows(X)` / `filter_index(X, idxs)`, same as `bootstrap()`. It passed CI because no `store_bootstraps` test uses a DataFrame `X` — worth adding a `store_bootstraps=True` → `predict(return_ci=True)` case with both pandas and polars `X` (and a DR-classifier polars case, still untested).

### Non-blocking
- `slearner.py:138-139` (and classifier `:385-386`): `prepend_column(0.0/1.0, X)` is rebuilt inside the per-group loop though it only depends on `X` — 2·N copies instead of 2. Hoist above the loop (this was the earlier "build once" optimization).
- `xlearner.py:161` (+ `BaseXClassifier`): the outcome model is fit with `filter_mask(y_filt, w == 1)`, but `y_filt_np = to_numpy(y_filt)` is computed one line below — use `y_filt_np[w == 1]` to drop the extra filter and avoid handing a native Series to `model.fit()` as the target.
- `tests/test_polars_support.py:33`: use `RANDOM_SEED` from `tests/const.py` rather than a hardcoded `RANDOM_STATE = 42` (repo convention).

Happy to re-review quickly once the bootstrap path is fixed.

jeongyoonlee (approved):
Blocker resolved — `_fit_bootstrap_clone` now routes through `n_rows(X)` / `filter_index(X, idxs)`, matching `bootstrap()`, so the `fit(store_bootstraps=True)` → `predict(return_ci=True)` path no longer does `X[idxs]` on a native `X` (which was a `KeyError` on pandas and a failure on polars). Confirmed no residual bare `X[...]` indexing remains anywhere in `base.py` or the learners.

Approving. The remaining items are all non-blocking polish + a test-coverage gap — filed as #920 so they don't hold up the merge:
- S-learner `predict` still rebuilds `prepend_column(0.0/1.0, X)` inside the per-group loop (hoist it)
- X-learner fit target: `y_filt_np[w == 1]` instead of re-`filter_mask(y_filt, w == 1)`
- test: use `RANDOM_SEED` from `tests/const.py`
- add a `store_bootstraps=True` → `predict(return_ci=True)` test with a DataFrame `X` (the path just fixed)

Thanks for the thorough iteration on this, @aman-coder03 — great to have native Polars support land.

PR #910, opened 2026-06-11
Fix #904: Prevent deepcopy of fitted templates in bootstrap and correct predict validation ordering

### LINKED ISSUE #904: Follow-ups from #886: bootstrap clone(safe=False) deepcopies fitted models; predict() validation ordering
Two non-blocking follow-ups identified while reviewing #886 (now merged). Both verified against current `master`.

1. Bootstrap `clone(learner_template, safe=False)` deepcopies the *fitted* learner

`_fit_bootstrap_clone()` (`causalml/inference/meta/base.py:36`) does:

```python
learner_b = clone(learner_template, safe=False)
learner_b.fit(X=X_b, treatment=treatment_b, y=y_b, p=p_b)
```

and is invoked with `self` — a *fitted* learner — as the template (`base.py:141`, via `fit_bootstrap_ensemble`).

`BaseLearner` uses `metaclass=ABCMeta` and does **not** inherit `sklearn.BaseEstimator` or define `get_params`, so `sklearn.base.clone(..., safe=False)` falls back to `copy.deepcopy()`. That deepcopies the **fitted** estimator (its `models_c`/`models_t` and learned weights) on every bootstrap iteration, only to overwrite them in the subsequent `.fit()`. This is wasted work, and for `warm_start=True` base learners it can change results (the clone starts from fitted weights rather than a fresh estimator).

The docstring states this "follows EconML's BootstrapEstimator pattern," which clones the **unfitted** estimator (parameters only).

**Suggested fix:** deepcopy unfitted templates rather than the fitted learner — stash `self._model_c_template = self.model_c` (and `_model_t_template`) in `__init__` and build each bootstrap clone from those, mirroring `BaseLearner.bootstrap()`.

2. `return_ci + return_components` validation ordering differs between regressor and classifier

- `BaseTLearner.predict()` raises the guard at the top (`tlearner.py:189`), before any computation.
- `BaseTClassifier.predict()` raises the same `ValueError` only at `tlearner.py:493` — **after** computing `model_c.predict_proba(X)` (`:470`) and the per-group `predict_proba` loop (`:474`).

Harmless but inconsistent and wasteful. Move the classifier check to the top of `predict()` to match the regressor and fail fast.

---

Context: follow-ups from #886.

cc @aman-coder03 (original contributor) — good follow-up if you'd like to pick it up.

Saurav-Gupta-9741:
## Description
This PR addresses the two follow-up items identified in [Issue #904](https://github.com/uber/causalml/issues/904). 

### 1. `predict()` Validation Ordering
In `BaseTClassifier.predict()`, the validation check for `return_ci` and `return_components` was being raised *after* computing the predictive probabilities. 
* **Fix:** Moved the `ValueError` guard to the very top of `BaseTClassifier.predict()` to match `BaseTRegressor.predict()` and ensure the method fails fast before performing unnecessary computation.

### 2. Deepcopying Fitted Models in Bootstrap
During `fit_bootstrap_ensemble`, the system was passing the fully fitted `self` to `_fit_bootstrap_clone`, which caused `clone(safe=False)` to deepcopy the entire fitted estimator (including the heavy `models_c` and `models_t` arrays) on every bootstrap iteration. This wasted memory and caused inconsistencies for `warm_start=True` base learners.
* **Fix:** 
  * Stashed `self._model_t_template = self.model_t` in `BaseTLearner.__init__` (mirroring the control model).
  * Implemented a `_get_unfitted_template()` method inside `BaseLearner` that performs a shallow copy and systematically drops the heavy fitted arrays (`models_c`, `models_t`, `bootstrap_models_`, etc.) to create a lightweight template.
  * Updated `fit_bootstrap_ensemble` to build each bootstrap clone from this clean template, mirroring the pattern in `BaseLearner.bootstrap()`.

## Note to Reviewers
If you find anything that needs adjustment or if there is a preferred way you'd like the templates handled for other specific meta-learners, please let me know and I will fix the changes accordingly! 

Looking forward to your review and response. Thank you!

### DIFF
--- a/causalml/inference/meta/base.py
+++ b/causalml/inference/meta/base.py
@@ -98,6 +98,24 @@ class BaseLearner(metaclass=ABCMeta):
         self.fit(X=X_b, treatment=treatment_b, y=y_b, p=p_b)
         return self.predict(X=X, p=p)
 
+    def _get_unfitted_template(self):
+        """Create a lightweight, unfitted copy of the meta-learner to prevent deepcopying heavy fitted weights."""
+        import copy
+        template = copy.copy(self)
+        
+        # Remove heavy fitted arrays/models
+        for attr in ["models", "models_c", "models_t", "bootstrap_models_", "propensity_model", "propensity", "models_tau"]:
+            if hasattr(template, attr):
+                delattr(template, attr)
+                
+        # Revert model_c and model_t to their unfitted state
+        if hasattr(template, "_model_c_template"):
+            template.model_c = template._model_c_template
+        if hasattr(template, "_model_t_template"):
+            template.model_t = template._model_t_template
+            
+        return template
+
     def fit_bootstrap_ensemble(
         self,
         X,
@@ -137,8 +155,9 @@ class BaseLearner(metaclass=ABCMeta):
         seeds = rng.randint(0, np.iinfo(np.int32).max, size=n_bootstraps)
         logger.info("Storing bootstrap ensemble ({} iterations)".format(n_bootstraps))
 
+        learner_template = self._get_unfitted_template() if hasattr(self, '_get_unfitted_template') else self
         self.bootstrap_models_ = Parallel(n_jobs=n_jobs)(
-            delayed(_fit_bootstrap_clone)(self, X, treatment, y, p, s, bootstrap_size)
+            delayed(_fit_bootstrap_clone)(learner_template, X, treatment, y, p, s, bootstrap_size)
             for s in tqdm(seeds)
         )
 
--- a/causalml/inference/meta/base.py
+++ b/causalml/inference/meta/base.py
@@ -101,19 +101,28 @@ class BaseLearner(metaclass=ABCMeta):
     def _get_unfitted_template(self):
         """Create a lightweight, unfitted copy of the meta-learner to prevent deepcopying heavy fitted weights."""
         import copy
+
         template = copy.copy(self)
-        
+
         # Remove heavy fitted arrays/models
-        for attr in ["models", "models_c", "models_t", "bootstrap_models_", "propensity_model", "propensity", "models_tau"]:
+        for attr in [
+            "models",
+            "models_c",
+            "models_t",
+            "bootstrap_models_",
+            "propensity_model",
+            "propensity",
+            "models_tau",
+        ]:
             if hasattr(template, attr):
                 delattr(template, attr)
-                
+
         # Revert model_c and model_t to their unfitted state
         if hasattr(template, "_model_c_template"):
             template.model_c = template._model_c_template
         if hasattr(template, "_model_t_template"):
             template.model_t = template._model_t_template
-            
+
         return template
 
     def fit_bootstrap_ensemble(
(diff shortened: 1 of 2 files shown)


Saurav-Gupta-9741:
Hi @jeongyoonlee Sir and @raman-coder03 Sir,

I've gone ahead and submitted the fixes for the two follow-ups we discussed in #904. I've implemented the _get_unfitted_template() logic to prevent the heavy deepcopy memory leak during bootstrap, and I've also moved the predict() validation guard to fail-fast as requested.

I noticed the CI workflows are currently paused awaiting maintainer approval (since this is my first contribution here). Whenever you have a moment to click "Approve and run," the tests should execute perfectly! Also, I have already run the black formatter locally to ensure it passes the linting checks.

Please let me know if you'd like any adjustments or if you have any feedback on the implementation. Thank you for your time!

jeongyoonlee (changes requested):
Thanks for picking this up

One change requested before merge:

### Keep `BaseLearner` learner-agnostic

`_get_unfitted_template()` is defined on `BaseLearner`, but it encodes T-learner specifics (resets `model_c`/`model_t`, deletes `models_c`/`models_t`). The base class shouldn't know about any one learner's model attributes. Two concrete problems:

- It only **resets** `model_c`/`model_t`. If `store_bootstraps` is later wired into S/X/R/DR learners, the template will silently carry **fitted** sub-models (`model`, `model_mu_*`, `model_tau_*`) → warm-start carryover with no error.
- The generic-looking delete list (`models`, `models_tau`, `propensity`, …) is dead code for T-learner and misleading about the method's generality.

Please replace it with a small overridable seam so each learner owns its own reset:

```python
# base.py  (add `import copy` at module top)
def _unfitted_clone(self):
    """Return an unfitted copy for bootstrap refitting. Subclasses that hold fitted
    sub-models should override to reset them to their unfitted templates."""
    return clone(self, safe=False)        # default = prior behavior
```

```python
# tlearner.py  (BaseTLearner)
def _unfitted_clone(self):
    template = copy.copy(self)
    for attr in ("models_c", "models_t", "bootstrap_models_"):
        if hasattr(template, attr):
            delattr(template, attr)
    template.model_c = self._model_c_template
    template.model_t = self._model_t_template
    return template
```

```python
# fit_bootstrap_ensemble: drop the always-True hasattr guard
learner_template = self._unfitted_clone()
```

This keeps `model_c`/`model_t` knowledge in `tlearner.py` (where they're defined), co-located with `_model_c_template`/`_model_t_template`.

While you're in that hunk: `_fit_bootstrap_clone`'s docstring still says *"the fitted learner to clone as a template"* — it's now an unfitted template, so please update it.

### Note (no action needed here)

The canonical fix is to make the meta-learners proper scikit-learn estimators, so `clone(self)` yields an unfitted template for free and this whole template-stripping mechanism goes away. That's a cross-cutting refactor of every learner's `__init__`, so it's **out of scope for this PR** — I'll file a separate tracking issue for it. This PR's localized fix is the right thing for now.

Saurav-Gupta-9741:
Respected @jeongyoonlee Sir,

I have updated the code and refactored the unfitted cloning logic to be completely learner-agnostic, exactly as you requested. I truly appreciate your time and the detailed feedback you provided, as it was a great learning experience for me.

Please let me know if there are any further changes required before merging.

Thank you once again for your valuable guidance!

jeongyoonlee (approved):
LGTM

PR #914, opened 2026-06-17
Make Cython line tracing opt-in to keep release wheels fast

### LINKED ISSUE #913: CausalRandomForestRegressor gets slower when you raise `n_jobs` (the hardcoded Cython `linetrace` directive compiles profiling hooks into the released wheels)
I was benchmarking `CausalRandomForestRegressor` and noticed it was slow, so I
did the obvious thing and set `n_jobs` to use more cores. It got slower. Setting
`n_jobs=4` was several times slower than the default. That seemed backwards, so I
dug into it, and the cause turned out to be a Cython build directive that ships
in the wheels.

## What I saw

On the stock PyPI wheel (0.16.0), 50 trees, 12k rows:

| setting                  |    time |
|--------------------------|--------:|
| default (`n_jobs` unset) |  29.0 s |
| `n_jobs=1`               |  29.5 s |
| `n_jobs=4`               | 192.6 s |

So the one knob you'd reach for to speed things up makes it about 6x slower
instead. There's no error or warning. If you scale up (say `n_estimators=100,
n_jobs=-1`) a fit that should take seconds can run for tens of minutes, and it
just looks like the model is heavy.

## The cause, and a check on it

The five tree Cython sources hardcode `# cython: linetrace=True` (it's handy for
measuring Cython coverage in tests). With that directive set, cythonize emits the
profiling/tracing machinery and, on CPython, compiles it in by default: the
generated C defines `CYTHON_PROFILE 1`, which gates the `__Pyx_TraceCall` /
`__Pyx_TraceSetupAndCall` hooks into every function, including the `nogil`
tree-building criterion functions in the hot loop.

To be precise about what is and isn't on: full line tracing (`CYTHON_TRACE`)
would need `CYTHON_TRACE_NOGIL=1` passed at build time, and that isn't set, so
`CYTHON_TRACE` stays 0. But the profiling path (`CYTHON_PROFILE`) is enough on
its own. Those hooks touch the GIL, so when the forest fits with
`prefer="threads"` the worker threads serialize on the GIL instead of running in
parallel. More threads means more contention, which is why `n_jobs=4` is the
slowest row in the table.

I checked this by building from source and toggling only that one directive,
nothing else (same Cython, same machine, same `setup.py`):

| build (only the directive differs) | profiling hooks in `.so` | n_jobs=1 | n_jobs=4 |
|------------------------------------|:------------------------:|---------:|---------:|
| master as-is (`linetrace=True`)    | yes (2 symbols)          |  29.4 s  | 190.9 s  |
| directive removed                  | none                     |   4.1 s  |   1.7 s  |

The first row matches the PyPI wheel. Dropping the directive removes the hooks
from the binary, makes the single-threaded fit ~7x faster, and turns `n_jobs`
back into a speedup (n_jobs=4 goes from slowest to fastest). For reference, the
hooks show up in the shipped wheel too:

```console
$ nm .../causalml/inference/tree/causal/_criterion.cpython-312-darwin.so \
    | grep -E "TraceSetupAndCall|call_return_trace"
... __Pyx_TraceSetupAndCall ...
... __Pyx_call_return_trace_func ...
```

The directive is in 5 `.pyx` files (and the matching `.pxd` headers):

```console
$ grep -rl "linetrace" causalml/ --include="*.pyx" --include="*.pxd"
causalml/inference/tree/_tree/_tree.pyx
causalml/inference/tree/_tree/_criterion.pyx
causalml/inference/tree/_tree/_splitter.pyx
causalml/inference/tree/causal/_builder.pyx
causalml/inference/tree/causal/_criterion.pyx
... plus the .pxd headers
```

## Reproducing it with Docker

If you want to check it on a clean machine, these two files do it end to end:

```bash
docker build -t cf-repro . && docker run --rm cf-repro
```

It installs the published `causalml==0.16.0` wheel, checks whether the shipped
`.so` has the profiling hooks compiled in, then times the fit at `n_jobs` of
unset, 1, and 4.

<details><summary><code>Dockerfile</code></summary>

```dockerfile
# Pinned to linux/amd64 on purpose. The problem is in the prebuilt wheels, and
# there's no 0.16.0 aarch64 wheel, so on ARM hosts pip builds from source
# (which doesn't get instrumented) and you won't see it. amd64 is what most CI
# and servers run anyway. On Apple Silicon this runs under emulation, slower but
# it still reproduces.
FROM --platform=linux/amd64 python:3.12-slim
RUN apt-get update && apt-get install -y --no-install-recommends libgomp1 \
    && rm -rf /var/lib/apt/lists/*
RUN pip install --no-cache-dir "causalml==0.16.0" "scikit-learn==1.6.1"
COPY repro.py /repro.py
CMD ["python", "/repro.py"]
```
</details>

<details><summary><code>repro.py</code></summary>

```python
import os, time
from importlib.metadata import version
import numpy as np
import causalml
from causalml.dataset import make_uplift_classification
from causalml.inference.tree import CausalRandomForestRegressor
from causalml.inference.tree.causal import _criterion
from sklearn.model_selection import train_test_split

TREES = int(os.environ.get("TREES", "40"))
print(f"causalml      : {version('causalml')}")
print(f"CPU cores     : {os.cpu_count()}")

# Check the shipped binary directly, no nm/binutils needed. These symbols are
# Cython's profile/trace hooks; they're only compiled in when the linetrace
# (or profile) directive is active.
blob = open(_criterion.__file__, "rb").read()
hooked = b"__Pyx_TraceSetupAndCall" in blob or b"__Pyx_call_return_trace_func" in blob
print(f"profile hooks : {'present (instrumented)' if hooked else 'absent'}")

df, feats = make_uplift_classification(
    n_samples=10000, treatment_name=["control", "treatment1"], y_name="conversion",
    positive_class_proportion=0.30, n_uplift_increase_dict={"treatment1": 3},
    delta_uplift_increase_dict={"treatment1": 0.10},
    n_uplift_increase_mix_informative_dict={"treatment1": 1}, random_seed=42)
df["is_sent"] = (df["treatment_group_key"] == "treatment1").astype(int)
X, w, y = df[feats].values, df["is_sent"].values, df["conversion"].values
idx, _ = train_test_split(np.arange(len(df)), test_size=0.4, random_state=42)
Xtr, wtr, ytr = X[idx], w[idx], y[idx]

for label, kw in [("default", {}), ("n_jobs=1", {"n_jobs": 1}), ("n_jobs=4", {"n_jobs": 4})]:
    t = time.time()
    CausalRandomForestRegressor(n_estimators=TREES, control_name=0,
                                random_state=42, **kw).fit(Xtr, wtr, ytr)
    print(f"  {label:10s}: {time.time()-t:6.1f}s")
```
</details>

Output on my machine (amd64 wheel, 20 trees):

```
profile hooks : present (instrumented)  [_criterion.cpython-312-x86_64-linux-gnu.so]
  default   :   26.2s
  n_jobs=1  :   26.2s
  n_jobs=4  :  147.0s
```

The x86_64 Linux wheel has the hooks too, so this isn't macOS-specific.

## What a fix looks like

Don't hardcode the directive in the sources. Gate it behind an env var so
coverage builds can still turn it on, but release wheels don't get it. In
`setup.py`:

```python
import os
linetrace = os.environ.get("CYTHON_TRACE", "0") not in ("0", "", "false", "False")
define_macros = [("CYTHON_TRACE", "1"), ("CYTHON_TRACE_NOGIL", "1")] if linetrace else []
# Extension(..., define_macros=define_macros)
# cythonize(..., compiler_directives={"linetrace": linetrace})
```

Then remove `# cython: linetrace=True` from the 5 `.pyx` files and their `.pxd`
headers. Coverage runs build with `CYTHON_TRACE=1 python setup.py build_ext`, and
everything else stays uninstrumented.

I tried this locally (the A/B above is exactly that change). Single-threaded fit
drops from ~29 s to ~4 s, `n_jobs=4` goes from slowest to fastest, predictions
are unchanged (a downstream T-Learner Qini score matched to 6 decimal places),
and `tests/test_causal_trees.py` still passes.

Happy to send a PR if this looks right.

## Environment
- causalml 0.16.0 (PyPI wheel)
- Python 3.12, scikit-learn 1.6.1
- Reproduced on both macOS (Apple Silicon) and the amd64 Linux wheel

HSJung93:
## Proposed changes

Fixes #913.

The 5 tree Cython sources hardcode # cython: linetrace=True (in the .pyx files and their .pxd headers), which compiles Cython's profiling hooks into the released wheels: on CPython the generated C sets CYTHON_PROFILE=1, inserting __Pyx_TraceCall/__Pyx_TraceSetupAndCall into the nogil tree-building loop. Those hooks take the GIL, so fitting CausalRandomForestRegressor with prefer="threads" serializes the workers and raising n_jobs makes it slower. (Line tracing, CYTHON_TRACE, would also need CYTHON_TRACE_NOGIL=1, which isn't set — the profiling path alone is enough.)

Confirmed by building from source and toggling only this one directive (same Cython, same machine): single-threaded fit 29.4s → 4.1s, n_jobs=4 190.9s → 1.7s. Numerical output unchanged (downstream T-Learner Qini matched to 6 d.p.); full suite passes.

The fix removes the directive from the .pyx/.pxd files and makes it opt-in via the CYTHON_TRACE env var, so coverage builds can still enable it.

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/causalml/inference/tree/_tree/_criterion.pxd
+++ b/causalml/inference/tree/_tree/_criterion.pxd
@@ -11,7 +11,6 @@
 # cython: boundscheck=False
 # cython: wraparound=False
 # cython: language_level=3
-# cython: linetrace=True
 
 # See _criterion.pyx for implementation details.
 from ._typedefs cimport float64_t, int8_t, int32_t, intp_t
--- a/causalml/inference/tree/_tree/_criterion.pyx
+++ b/causalml/inference/tree/_tree/_criterion.pyx
@@ -16,7 +16,6 @@
 # cython: boundscheck=False
 # cython: wraparound=False
 # cython: language_level=3
-# cython: linetrace=True
 
 from libc.string cimport memcpy
 from libc.string cimport memset
--- a/causalml/inference/tree/_tree/_splitter.pxd
+++ b/causalml/inference/tree/_tree/_splitter.pxd
@@ -12,7 +12,6 @@
 # cython: boundscheck=False
 # cython: wraparound=False
 # cython: language_level=3
-# cython: linetrace=True
 
 # See _splitter.pyx for details.
 from ._criterion cimport Criterion
--- a/causalml/inference/tree/_tree/_splitter.pyx
+++ b/causalml/inference/tree/_tree/_splitter.pyx
@@ -15,7 +15,6 @@
 # cython: boundscheck=False
 # cython: wraparound=False
 # cython: language_level=3
-# cython: linetrace=True
 
 from cython cimport final
 from libc.math cimport isnan
--- a/causalml/inference/tree/_tree/_tree.pxd
+++ b/causalml/inference/tree/_tree/_tree.pxd
@@ -13,7 +13,6 @@
 # cython: boundscheck=False
 # cython: wraparound=False
 # cython: language_level=3
-# cython: linetrace=True
 
 # See _tree.pyx for details.
 
--- a/causalml/inference/tree/_tree/_tree.pyx
+++ b/causalml/inference/tree/_tree/_tree.pyx
@@ -17,7 +17,6 @@
 # cython: boundscheck=False
 # cython: wraparound=False
 # cython: language_level=3
-# cython: linetrace=True
 
 from cpython cimport Py_INCREF, PyObject, PyTypeObject
 
--- a/causalml/inference/tree/_tree/_utils.pxd
+++ b/causalml/inference/tree/_tree/_utils.pxd
@@ -11,7 +11,6 @@
 # cython: boundscheck=False
 # cython: wraparound=False
 # cython: language_level=3
-# cython: linetrace=True
 
 # See _utils.pyx for details.
 
--- a/causalml/inference/tree/causal/_builder.pxd
+++ b/causalml/inference/tree/causal/_builder.pxd
@@ -2,7 +2,6 @@
 # cython: boundscheck=False
 # cython: wraparound=False
 # cython: language_level=3
-# cython: linetrace=True
 
 from .._tree._tree cimport Node, Tree, TreeBuilder
 from .._tree._splitter cimport Splitter, SplitRecord
--- a/causalml/inference/tree/causal/_builder.pyx
+++ b/causalml/inference/tree/causal/_builder.pyx
@@ -3,7 +3,6 @@
 # cython: boundscheck=False
 # cython: wraparound=False
 # cython: language_level=3
-# cython: linetrace=True
 
 
 from libc.stdint cimport INTPTR_MAX
--- a/causalml/inference/tree/causal/_criterion.pxd
+++ b/causalml/inference/tree/causal/_criterion.pxd
@@ -2,7 +2,6 @@
 # cython: boundscheck=False
 # cython: wraparound=False
 # cython: language_level=3
-# cython: linetrace=True
 # distutils: language = c++
 
 from libc.math cimport fabs
--- a/causalml/inference/tree/causal/_criterion.pyx
+++ b/causalml/inference/tree/causal/_criterion.pyx
@@ -2,7 +2,6 @@
 # cython: boundscheck=False
 # cython: wraparound=False
 # cython: language_level=3
-# cython: linetrace=True
 
 
 cdef int32_t CONTROL_GROUP_IDX = 0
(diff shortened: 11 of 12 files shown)


HSJung93:
"This is a build-flag change (line tracing), not a behavioral change, so I didn't add a new test. Existing tests/test_causal_trees.py passes and numerical output is unchanged."

jeongyoonlee (approved):
Good catch! Thanks for your contribution, @HSJung93

PR #918, opened 2026-06-19
Add JAX/flax.nnx backend for DragonNet

### LINKED ISSUE #917: Add JAX/flax.nnx backend for DragonNet

CausalML currently offers DragonNet only through a TensorFlow backend. As JAX emerges as a standard in ML research, with a familiar NumPy-compatible API, native JIT compilation, and a growing ecosystem (Flax, Optax, Orbax), practitioners who work in JAX-based stacks have no native path to use DragonNet without adopting a different framework. Adding a JAX backend would give developers a modern, performant alternative that fits naturally into their existing workflows.

A JAX backend for DragonNet under `causalml/inference/jax/`, using `flax.nnx` as the neural network framework and `optax` for optimization. The new module would expose an identical sklearn-style API (`fit`, `predict`, `predict_tau`, `fit_predict`, `save`, `load`) and be available as an opt-in extra (`pip install "causalml[jax]"`), following the same pattern as the existing `tf` and `torch` extras.

Keeping TensorFlow as the only backend. However, given JAX's growing adoption in academic research and its performance advantages via JIT compilation, a dedicated backend seems more appropriate than a wrapper solution.

**Additional context**

- Original DragonNet paper: Shi, Blei & Veitch (2019) — https://arxiv.org/pdf/1906.02120.pdf
- flax.nnx docs: https://flax.readthedocs.io/en/latest/nnx/index.html
- JAX ecosystem: https://docs.jaxstack.ai/en/latest/index.html
- Existing TF implementation: `causalml/inference/tf/dragonnet.py`

xrhd:
## Proposed changes

Adds a JAX backend for DragonNet under `causalml/inference/jax/`, using `flax.nnx`
as the neural network framework and `optax` for optimization. Exposes an identical
sklearn-style API (`fit`, `predict`, `predict_tau`, `fit_predict`, `save`, `load`)
and follows the same module pattern as the existing `tf` and `torch` backends.

Resolves uber/causalml#917

## Types of changes

## Checklist

## Changes

- `causalml/inference/jax/__init__.py` — new module, exports `DragonNet`
- `causalml/inference/jax/utils.py` — pure-JAX loss functions (`regression_loss`, `binary_classification_loss`, `dragonnet_loss_binarycross`, `make_tarreg_loss`)
- `causalml/inference/jax/dragonnet.py` — `EpsilonLayer`, `DragonNetModule` and `DragonNet` sklearn wrapper using `flax.nnx` + `optax`. Serialization via `orbax-checkpoint`
- `tests/test_jax_dragonnet.py` — fit/predict + AUUC check, propensity scores, save/load round-trip
- `tests/conftest.py` — `--runjax` option and `jax` marker (same pattern as `--runtf`/`--runtorch`)
- `pyproject.toml` — `[jax]` optional-dependencies group (opt-in, same as `tf` and `torch`)
- `docs/examples/dragonnet_jax_vs_tf.ipynb` — benchmark notebook comparing JAX vs TF on the IHDP dataset (ATE, MAE, AUUC, training time)

## Tests

```bash
pip install ".[jax,test]"
pytest tests/test_jax_dragonnet.py --runjax -v
```

## Dependencies

```toml
jax = [
    "jax>=0.4.1",
    "flax>=0.10.0",
    "optax>=0.1.4",
    "orbax-checkpoint>=0.10.0",
]
```

## Further comments

The JAX implementation mirrors the TF DragonNet architecture exactly (3 shared representation layers, propensity head, two outcome heads, EpsilonLayer). The two-phase training (Adam warm-up + SGD Nesterov) is replicated using `optax`. The main design decision was to keep the dependency opt-in via `causalml[jax]` to avoid impacting users who do not need the JAX backend.

## References

- Shi, Blei & Veitch (2019) — https://arxiv.org/pdf/1906.02120.pdf
- flax.nnx: https://flax.readthedocs.io/en/latest/nnx/index.html
- JAX ecosystem: https://docs.jaxstack.ai/en/latest/index.html


jeongyoonlee (changes requested):
Thanks for your contribution, @xrhd. A few things need addressing before we can merge.

## Blocking

**1. flax API vs the pinned floor.** The optimizer calls (`nnx.Optimizer(model, tx, wrt=nnx.Param)`, `update(model, grads)`) are the flax 0.11 API. `flax>=0.10.0` permits 0.10.x, where these signatures don't exist, so a clean install fails with `TypeError`. Please bump the pin to `flax>=0.11.0`. Also add a `--runjax` CI lane (mirroring tf/torch) — nothing here runs in CI today.

**2. Default `batch_size=256` can silently train zero steps.** In `_run_training_loop`, `n_full = (n_train // batch_size) * batch_size` is `0` when `n_train < batch_size`, so the batch loop never runs and `fit()` returns an untrained model with no error. At the default 256 + `val_split=0.2`, any dataset under ~320 rows hits this — DragonNet's typical regime (IHDP ≈ 747). Please clamp the batch size to `n_train` (or raise on `n_full == 0`) and default to 64 to match TF. The "batch_size=64" comment in that function is also stale.

## Non-blocking

- **`seed` ignored in shuffling.** `_run_training_loop` hardcodes `np.random.default_rng(0)`; pass `self.seed` through so it matches the docstring.
- **`load(path, input_dim)`** adds a required arg the tf/torch backends lack. Persist `input_dim`/`neurons_per_layer` in the checkpoint so `load` can reconstruct on its own.
- **`save()` path handling.** orbax wants an absolute path and errors if the dir exists — `os.path.abspath(...)` + `force=True` handles both.
- **No LR schedule / NaN guard** (TF has `ReduceLROnPlateau` + `TerminateOnNaN`). Fine to skip, but note it in the docstring.

xrhd:
Hi @jeongyoonlee, thanks for the thorough review. I've addressed all the blocking items and the non-blocking suggestions in commit `928903e`.

**Blocking**
1. Bumped `flax` pin to `>=0.11.0` in `pyproject.toml` and added a `--runjax` CI lane in `test-build-from-source.yml` that mirrors the existing `--runtf` lane.
2. Changed default `batch_size` to `64`, clamped the effective batch size to `n_train` inside `_run_training_loop` to avoid silently running zero steps, and removed the stale `batch_size=64` comment.

**Non-blocking**
- `_run_training_loop` now receives `self.seed` instead of hardcoding `np.random.default_rng(0)`.
- `save()` persists `input_dim`, `neurons_per_layer`, and `reg_l2` into a `config.json` sidecar in the checkpoint directory. `load(path)` now reconstructs the model on its own; `input_dim` remains an optional fallback for older checkpoints.
- `save()` now uses `os.path.abspath(path)` and passes `force=True` to the orbax checkpointer.
- Added a docstring note that the JAX backend does not currently implement `ReduceLROnPlateau` or `TerminateOnNaN`.

**Validation**
- `pytest -vs tests/test_jax_dragonnet.py --runjax`
- `pytest -vs tests/ --runjax`
- `jupyter execute --inplace docs/examples/dragonnet_jax_vs_tf.ipynb`

Please take another look when you have a chance.

jeongyoonlee (approved):
Thanks for the follow-up, @xrhd. Both blockers from the previous review are resolved — the `flax>=0.11.0` pin now matches the optimizer API and a `--runjax` CI lane is wired up, and the `effective_batch_size = min(batch_size, n_train)` clamp removes the silent zero-step training. Loss math, L2 penalty, and EpsilonLayer init still match the TF reference. Approving.

Non-blocking (this PR or a follow-up):
- `docs/examples.rst` toctree omits the new `examples/dragonnet_jax_vs_tf` notebook, so Sphinx won't render it — add the entry.
- `docs/installation.rst` lists only the `tf`/`torch` extras for DragonNet; add a line for `pip install causalml[jax]`.

xrhd:
Thanks again for the review and for approving, @jeongyoonlee!

I noted the two documentation follow-ups:
- Adding the `examples/dragonnet_jax_vs_tf` notebook to `docs/examples.rst`.
- Adding `pip install causalml[jax]` to `docs/installation.rst`.

I'd like to handle those in a separate follow-up PR dedicated to documentation, where I can also improve the examples and add more guidance around the JAX backend. Please let me know if that works for you.

jeongyoonlee:
Sounds good. Once the build completes successfully, we will merge and close this PR. Thanks, @xrhd!

PR #923, opened 2026-07-03
Add `return_components` support to R-learner

### LINKED ISSUE #304: return_components for R-Learner
I wanted to receive to components of the CATE prediction via R-Learner. Unfortunately, it is not possible to retrieve this information because it is not defined [here](https://github.com/uber/causalml/blob/master/causalml/inference/meta/rlearner.py#L122-L137). I tried to solve it on my own but I was not successful because it is unclear which "predict"-function is used in Line 134 - `dhat = self.models_tau[group].predict(X)`. Which predict is used for calculation?

It would be very useful to conduct a validation of all Meta Learners in the [here proposed way](https://github.com/uber/causalml/issues/290).

aman-coder03:
## Proposed changes
this PR adds `return_components` support to the R-Learner, bringing its API in line with the existing T- and X-Learner implementations

specific changes..
- adds a `return_components` argument to `predict()` and `fit_predict()` for both `BaseRLearner` and `BaseRClassifier`
- returns the nuisance components used by the R-Learner
  - `yhat`: outcome model predictions (`E[Y|X]`)
  - `p`: propensity score estimates (`E[W|X]`)
- adds the same mutual exclusion guard as other meta-learners, preventing `return_ci` and `return_components` from being used together
- fits the nuisance outcome model after cross-validation so that it can be used for inference-time component retrieval.
- adds tests covering the new `return_components` functionality for both `predict()` and `fit_predict()` along with the mutual exclusion behavior

fixes #304 

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/causalml/dataset/synthetic.py
+++ b/causalml/dataset/synthetic.py
@@ -411,11 +411,11 @@ def get_synthetic_preds_holdout(
                 learner.fit(X=X_train, p=p_hat_train, treatment=w_train, y=y_train)
 
                 preds_dict_train["{} Learner ({})".format(label_l, label_m)] = (
-                    learner.predict(X=X_train)
+                    learner.predict(X=X_train).flatten()
                 )
 
                 preds_dict_valid["{} Learner ({})".format(label_l, label_m)] = (
-                    learner.predict(X=X_val)
+                    learner.predict(X=X_val).flatten()
                 )
 
     return preds_dict_train, preds_dict_valid
--- a/causalml/dataset/synthetic.py
+++ b/causalml/dataset/synthetic.py
@@ -409,11 +409,19 @@ def get_synthetic_preds_holdout(
             else:
                 learner = base_learner(model())
                 learner.fit(X=X_train, p=p_hat_train, treatment=w_train, y=y_train)
+
                 preds_dict_train["{} Learner ({})".format(label_l, label_m)] = (
-                    learner.predict(X=X_train).flatten()
+                    learner.predict(
+                        X=X_train,
+                        p=p_hat_train,
+                    ).flatten()
                 )
+
                 preds_dict_valid["{} Learner ({})".format(label_l, label_m)] = (
-                    learner.predict(X=X_val).flatten()
+                    learner.predict(
+                        X=X_val,
+                        p=p_hat_val,
+                    ).flatten()
                 )
 
     return preds_dict_train, preds_dict_valid
--- a/causalml/dataset/synthetic.py
+++ b/causalml/dataset/synthetic.py
@@ -411,17 +411,11 @@ def get_synthetic_preds_holdout(
                 learner.fit(X=X_train, p=p_hat_train, treatment=w_train, y=y_train)
 
                 preds_dict_train["{} Learner ({})".format(label_l, label_m)] = (
-                    learner.predict(
-                        X=X_train,
-                        p=p_hat_train,
-                    ).flatten()
+                    learner.predict(X=X_train)
                 )
 
                 preds_dict_valid["{} Learner ({})".format(label_l, label_m)] = (
-                    learner.predict(
-                        X=X_val,
-                        p=p_hat_val,
-                    ).flatten()
+                    learner.predict(X=X_val)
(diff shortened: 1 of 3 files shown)


jeongyoonlee (changes requested):
Two blockers before merge:

**B1 — the `p` component is the stored training propensity, not recomputed for the passed `X`.**
In both `predict()` bodies, when `p is None` you set `p = self.propensity` (training-data propensity) while `yhat = self.model_mu.predict(X)` is computed for the passed `X`. The two returned components end up on different bases, and on a differently-sized `X` the `p` array length won't match `yhat`/`te` — with no error raised:

```python
rl.fit(X[:800], treatment[:800], y[:800], p=p[:800])
te, yhat, p_hat = rl.predict(X[800:1000], return_components=True)
# te: (200, 1), yhat: len 200, p_hat: len 800
```

Please mirror the X-learner, which recomputes propensity for `X` (xlearner.py:203/:643):
```python
if p is None:
    p = {g: self.propensity_model[g].predict(X) for g in self.t_groups}
else:
    p = self._format_p(p, self.t_groups)
```
Note `self.propensity_model` only exists when `fit()` ran with `p=None`; if the user supplied `p` at fit and then calls `predict(p=None)` there is no model to recompute from — raise a clear error there rather than returning stale training values. The current test doesn't surface this because it always passes `p=p_scores` and predicts on the training `X` of equal length.

**B2 — `predict(..., return_ci=True)` is accepted but never implemented.**
The new `predict()` signature adds `return_ci=False`, but the body only uses it for the `return_ci`/`return_components` mutual-exclusion guard — it never computes CIs. So `te, lb, ub = rl.predict(X, return_ci=True)` raises `ValueError` at the unpack site (same footgun as #886). R-learner has no per-predict bootstrap path, so rather than accept-and-ignore, please drop `return_ci` from `predict()` (keep the guard in `fit_predict`, which does implement it) or raise `NotImplementedError` when `True`.

**Tests:** the new coverage only exercises `BaseRLearner`. Since `BaseRClassifier` and `XGBRRegressor` both had `fit()`/`predict()` modified, please add `return_components` tests for the classifier override and XGBR, plus the `p=None` predict path and a predict on a different-sized `X` (that last one guards B1).

Non-blocking notes to follow separately.

jeongyoonlee (changes requested):
One blocker:

**`predict(X)` (with `return_components` off) now raises on a learner fit with an explicit `p`.**
The nuisance computation was hoisted above the te-only early return, so `p` (and `yhat`) are computed on every `predict()` even when `return_components=False`. On master `predict(X, p=None)` never touches `p` — the CATE comes only from `models_tau` — so this is a regression:

```python
rl = BaseRRegressor(learner=LinearRegression(), control_name="control")
rl.fit(X, treatment, y, p=p)   # fit with explicit propensity
rl.predict(X)                  # master: returns te;  this PR: ValueError: No propensity model is available.
```

That's why `get_synthetic_preds_holdout` and three existing tests had to start passing `p=` to `predict()`. Please move the `p`/`yhat` computation inside `if return_components:` so the plain `predict(X) → te` path stays as on master (and skips the extra work):

```python
te = np.zeros((n_rows(X), self.t_groups.shape[0]))
for i, group in enumerate(self.t_groups):
    te[:, i] = self.models_tau[group].predict(X)
if not return_components:
    return te
if p is None:
    if not hasattr(self, "propensity_model"):
        raise ValueError("No propensity model is available. ...")
    p = {g: self.propensity_model[g].predict(X) for g in self.t_groups}
else:
    p = self._format_p(p, self.t_groups)
yhat = self.model_mu.predict(X)
return te, yhat, p
```

Same change in the `BaseRClassifier` override. With that, the `p=` additions in `synthetic.py` and the tests become unnecessary.

Non-blocking notes to follow separately.

jeongyoonlee (approved):
LGTM.

One non-blocking follow-up: `self.model_mu.fit(X, y_np)` runs on every `fit()` even when `return_components` is never used — worth gating or lazy-fitting so the extra full-data fit only happens when the components are actually requested.

PR #925, opened 2026-07-03
Add Sensitivity bounds for ATE via Marginal Sensitivity Model

### LINKED ISSUE #916: Sensitivity bounds for CATE via Marginal Sensitivity Model (MSM)
`causalml.metrics.sensitivity` already has a `Sensitivity` framework: placebo treatment, random/irrelevant confounder, subset-data refutation, and `SensitivitySelectionBias` (based on Blackwell 2014). These are all refutation-style checks. They tell you whether your estimate looks fragile, but none of them give you an actual quantitative bound on how far the true CATE/ATE could move if there's unobserved confounding.

`SensitivitySelectionBias` also currently logs this on init,
> "Only works for linear outcome models right now. Check back soon." (`causalml/metrics/sensitivity.py` line 396)

that's a real gap for a library built around ML-based meta-learners (S/T/X/R/DR learners). The one sensitivity method that gives a numeric bound doesn't actually support the outcome models causalml is designed around. I'm always frustrated when I fit an X-learner or DR-learner on observational-ish data and the only "how robust is this" tooling available is a refutation test, not an actual interval

adding sensitivity bounds based on the Marginal Sensitivity Model (MSM), originally Tan (2006), with closed-form sharp bounds for (C)ATE worked out by Zhao, Small & Bhattacharya (2019, JRSS-B) and later Dorn & Guo (2023, JASA). The NeurIPS 2023 paper by Frauen, Melnychuk & Feuerriegel ("Sharp Bounds for Generalized Causal Sensitivity Analysis," arXiv:2305.16988) shows the standard MSM is the binary-treatment special case of a more general framework, and that their bounds coincide with the Dorn-Guo result there

why MSM specifically...
- it's model-agnostic:, it only needs propensity scores and the fitted outcome regression, both of which every meta-learner in `causalml.inference.meta` already produces. No restriction to linear outcomes like the current `SensitivitySelectionBias`
- the bounds are closed-form and sharp for the binary-treatment case, so no grid search or extra optimization loop is needed, and no heavy new dependency (the general GMSM paper needs normalizing flows for the continuous/mediated case, but that's not necessary here)
- it uses one interpretable sensitivity parameter, Gamma >= 1, that bounds how much an unobserved confounder could be affecting treatment assignment. Gamma = 1 just recovers the original point estimate. This Gamma parameterization is basically the standard in the sensitivity-analysis literature at this point (it's what `causalsens` uses in R, and what most of the recent papers compare against)

rough API sketch, following the existing `Sensitivity` class conventions...
```python
from causalml.metrics.sensitivity import SensitivityMSM

sens = SensitivityMSM(
    df=df,
    inference_features=feature_names,
    p_col='propensity_score',
    treatment_col='treatment',
    outcome_col='y',
    learner=my_fitted_meta_learner,
)

bounds_df = sens.get_msm_bounds(gamma=[1.0, 1.5, 2.0, 3.0])
#    gamma  ate_lower  ate_upper
# 0    1.0      0.42       0.42
# 1    1.5      0.21       0.61
# 2    2.0      0.05       0.74
# 3    3.0     -0.18       0.91

sens.breakdown_gamma()  # smallest Gamma that flips the sign of the effect

sens.plot_msm_bounds(gamma_range=(1, 5))
```

this would sit alongside `SensitivityPlaceboTreatment`, `SensitivityRandomCause`, etc. as another `Sensitivity` subclass, so it fits the existing module shape instead of introducing a separate API

for a first pass i would keep scope tight: binary treatment only, (C)ATE only (not mediation or distributional effects, which need the heavier neural density estimation from the general GMSM paper), and the closed-form Dorn & Guo bounds rather than the older non-sharp percentile-bootstrap version from Zhao et al. Propensity scores are already a dependency via `ElasticNetPropensityModel`, so there's nothing new to add there either

- just fixing `SensitivitySelectionBias` to support non-linear models directly. Possible, but Blackwell's selection-bias approach is designed around linear regression coefficients in a fairly fundamental way, so this would mean a different method under the same class rather than a small patch
- implementing the full GMSM from Frauen et al. (mediation analysis, continuous treatments, distributional effects). More general, but it needs conditional normalizing flows for density estimation, which is a much bigger lift and a new dependency. Binary MSM covers the common case (causalml's existing learners are basically all binary/multi-treatment CATE) without that cost
- pointing users to an external package (e.g. the authors' own `SharpCausalSensitivity` repo) instead of adding this in-tree. Doesn't integrate with causalml's existing learner objects or propensity scores, so users would have to glue things together themselves

**Additional context**
i know the maintainers have mentioned elsewhere (#725) that causalml's primary focus is HTE estimation for randomized experiments rather than general observational causal inference, and that sensitivity tooling should be framed carefully. I would actually frame this addition the same way the existing `Sensitivity` module already is: not as a green light for observational use, but as another way to show users how fragile their estimate is, just with a number attached instead of a refutation heuristic

If this direction seems reasonable i would like to take a shot at implementing it. Could start with just `get_msm_bounds()` for ATE on `BaseSLearner`/`BaseTLearner` output and extend from there once the core bound is in.
Let me know if this scope makes sense or if you'd rather see it shaped differently before I put together a PR

aman-coder03:
## Proposed changes
adds `SensitivityMSM`, a new `Sensitivity` subclass implementing sensitivity bounds for the ATE under the Marginal Sensitivity Model (MSM). the existing `SensitivitySelectionBias` only works for linear outcome models, so there wasn't a way to get an actual numeric bound (rather than a refutation-style check) on how far the true ATE could move under unobserved confounding when using causalml's meta-learners (S/T/X/R/DR)

MSM only needs propensity scores + the fitted outcome regression, both of which the existing meta-learners already produce, so no new dependency and no restriction to linear models

implementation is the closed-form, elementwise-sharp bound based on Dorn & Guo (2023, JASA) and the closed-form expressions in Dorn, Guo & Kallus (2024, arXiv:2112.11449). exposed as `get_msm_bounds(gamma=[...])`, following the existing `Sensitivity` class conventions (`sensitivity_estimate()`, registration via `get_class_object`/`sensitivity_analysis()`)

scope for this PR, per discussion in the issue...binary treatment only, ATE only. `plot_msm_bounds()` and `breakdown_gamma()` are deferred as follow-ups

docstring makes clear this is fragility diagnostic tooling (like the rest of the sensitivity module), not a green light for observational causal inference, and that the Gamma (propensity odds-ratio) parameterization is a different quantity from the partial-R² robustness value used by EconML/DoWhy

closes #916 

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/causalml/metrics/sensitivity.py
+++ b/causalml/metrics/sensitivity.py
@@ -255,23 +255,65 @@ class Sensitivity:
 
         return summary_df
 
+    # Learners whose fit_predict(return_components=True) does NOT return
+    # potential-outcome regressions (mu0, mu1). X-learner returns two CATE
+    # estimates from its tau models instead (xlearner.py); R-learner has no
+    # outcome-regression decomposition at all. Both are rejected explicitly
+    # rather than silently misinterpreted.
+    _UNSUPPORTED_POTENTIAL_OUTCOME_LEARNERS = (
+        "BaseXLearner",
+        "BaseXRegressor",
+        "BaseXClassifier",
+        "BaseRLearner",
+        "BaseRRegressor",
+        "BaseRClassifier",
+    )
+
     def get_potential_outcome_predictions(self, X, p, treatment, y):
         """Return separate potential-outcome predictions mu1_hat, mu0_hat.
+
+        Only supported for S/T/DR-learner-style objects, whose
+        fit_predict(..., return_components=True) returns the fitted
+        outcome regressions (mu0_hat, mu1_hat) directly. X-learner and
+        R-learner are explicitly unsupported: X-learner's "components"
+        are two CATE estimates from its second-stage tau models, not
+        potential outcomes, and R-learner has no outcome-regression
+        decomposition to extract.
 
         Args:
             X, p, treatment, y: same as get_prediction()
         Returns:
             (tuple of np.array): (mu1_hat, mu0_hat)
+        Raises:
+            NotImplementedError: if the learner does not expose
+                potential-outcome regressions via return_components.
         """
         learner = self.learner
+        learner_name = type(learner).__name__
+
+        if learner_name in self._UNSUPPORTED_POTENTIAL_OUTCOME_LEARNERS:
+            raise NotImplementedError(
+                "SensitivityMSM does not support {} yet: it needs potential-"
+                "outcome regressions (mu0_hat, mu1_hat), which this learner's "
+                "return_components does not expose. Use an S-learner, "
+                "T-learner, or DR-learner instead.".format(learner_name)
+            )
+
         try:
             _, yhat_cs, yhat_ts = learner.fit_predict(
                 X=X, p=p, treatment=treatment, y=y, return_components=True
             )
         except TypeError:
-            _, yhat_cs, yhat_ts = learner.fit_predict(
-                X=X, treatment=treatment, y=y, return_components=True
-            )
+            try:
+                _, yhat_cs, yhat_ts = learner.fit_predict(
+                    X=X, treatment=treatment, y=y, return_components=True
+                )
+            except TypeError:
+                raise NotImplementedError(
+                    "SensitivityMSM could not extract potential-outcome "
+                    "predictions from {}: fit_predict() does not support "
+                    "return_components=True.".format(learner_name)
+                )
         # yhat_cs/yhat_ts are dicts keyed by treatment group; binary case → one group
         group = list(yhat_cs.keys())[0]
         return yhat_ts[group], yhat_cs[group]
@@ -667,6 +709,11 @@ class SensitivityMSM(Sensitivity):
         As with the rest of this module, this is fragility-diagnostic
         tooling, not a license for observational causal inference; see
         causalml's HTE-for-experiments framing (issue #725).
+
+        Supported learners: S-learner, T-learner, DR-learner (any learner
+        whose fit_predict(return_components=True) returns potential-outcome
+        regressions mu0_hat, mu1_hat). X-learner and R-learner are not yet
+        supported and will raise NotImplementedError.
     """
 
     def __init__(self, *args, gamma=None, **kwargs):
--- a/causalml/metrics/sensitivity.py
+++ b/causalml/metrics/sensitivity.py
@@ -76,6 +76,25 @@ def alignment_att(alpha, p, treatment):
     return adj
 
 
+def msm_propensity_bounds(p, gamma):
+    """Clipped propensity bounds under the Marginal Sensitivity Model.
+
+    Reference: Tan, Z. (2006); Dorn, J. & Guo, K. (2023); closed-form
+    expressions in Dorn, J., Guo, K. & Kallus, N. (2024),
+    "Doubly-Valid/Doubly-Sharp Sensitivity Analysis..." arXiv:2112.11449.
+
+    Args:
+        p (np.array): estimated propensity score vector, in (0, 1)
+        gamma (float): sensitivity parameter, Gamma >= 1
+
+    Returns:
+        (tuple of np.array): (p_lower, p_upper) bounds on the true propensity
+    """
+    p_lower = p / (p + gamma * (1 - p))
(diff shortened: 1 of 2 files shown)


jeongyoonlee (changes requested):
Two blockers before merge, both in how potential outcomes are pulled from the learner:

**B1 — `get_potential_outcome_predictions` misreads X-learner components (and the test uses the X-learner).**
The method assumes `fit_predict(return_components=True)` returns `(te, μ0, μ1)`. That holds for S/T/DR (`yhat_cs`=μ0, `yhat_ts`=μ1), but the **X-learner returns `(te, dhat_cs, dhat_ts)`** — two CATE estimates from the tau models, not potential outcomes (xlearner.py:214-215,244). Replaying the PR logic on `synthetic_data(mode=1, n=100000)`:
- X-learner: `mean(mu0) == mean(mu1) == 0.689`, so `mu1 - mu0 ≈ 0`; residuals `y - mu` are inflated → **Γ=2 width 1.56 vs S/T ~1.26 (~24% off)**.
- The Γ=1 point estimate still looks right (0.517 vs true 0.500) only because AIPW degenerates to IPW when `mu1≈mu0` — which is exactly why the test passes.

Please extract μ0/μ1 learner-aware (for the X-learner use the outcome models `models_mu_c`/`models_mu_t`), or restrict + validate to S/T/DR and move the test off `BaseXLearner`. Adding an assertion that the Γ=1 bound ≈ the naive/true ATE would catch this class of error.

**B2 — R-learner path raises an uncaught `TypeError`.**
The description lists S/T/X/R/DR, but `BaseRLearner.fit_predict()` accepts no `return_components`, and both the `try` and `except` branches pass `return_components=True`, so the fallback doesn't rescue it:
```
TypeError: BaseRLearner.fit_predict() got an unexpected keyword argument 'return_components'
```
Please either drop R from the supported set (and the docstring) or raise a clear `NotImplementedError` naming the learner.

Non-blocking notes (sharpness wording — this is the conservative ZSBB/mean-residual bound rather than the quantile-balanced *sharp* Dorn-Guo bound; test seeding; propensity clipping) to follow separately.

jeongyoonlee (approved):
LGTM — both blockers are resolved: X/R learners are now rejected up front with a clear `NotImplementedError`, and the test moved to `BaseTLearner` with the `Γ=1 ≈ tau.mean()` guard that catches the B1 class of error. A few non-blocking notes to follow separately. Thanks!

jeongyoonlee:
A few non-blocking follow-ups — all fine to defer to a later PR, this is approved as-is:

1. **A whitelist would fail-safe better than the blocklist.** `_UNSUPPORTED_POTENTIAL_OUTCOME_LEARNERS` denies X/R by exact class name; a learner type added later defaults to "supported" and could silently miscompute (the same class of bug as B1). Consider allow-listing the S/T/DR families instead, so an unknown learner raises rather than guesses.

2. **Cover S and DR in the test.** `test_SensitivityMSM` only exercises `BaseTLearner`. S and DR are correct-by-shape but untested — worth adding them so all three supported learners are locked in.

3. **Sharpness wording.** The docstring's "sharp elementwise … Dorn & Guo" is really the conservative ZSBB / mean-residual AIPW *outer* bound (it thresholds on `resid ≥ 0` relative to the conditional mean μ), not the quantile-balanced *sharp* Dorn-Guo (2023) bound. Either soften to "conservative MSM AIPW bound" or implement quantile balancing. (The odds-ratio math and the worst-casing direction are correct.)

4. **Seed the test data.** `synthetic_data(mode=1, n=100000)` is unseeded; use `tests/const.py::RANDOM_SEED` for determinism (stable at n=100k in practice, but matches the repo convention).

5. **Propensity clipping.** `resid_t / w_t_ub` and `resid_c / (1 - w_c_ub)` divide by propensity bounds, so p∈{0,1} → inf/nan. Clip, or document the (0,1) requirement.

6. **Minor double-fit.** Invoking via `sensitivity_analysis(["MSM"])` fits twice (base `summary` → `get_prediction`, then `get_msm_bounds`). Harmless, just noting.

PR #929, opened 2026-07-04
feat: add model serialization API with version metadata (#892)

### LINKED ISSUE #892: [Enhancement] Model Serialization & Persistence for Trained Learners via save/load API
`causalML` has no standard way to persist a trained learner. Once you call `.fit()`, the only way to reuse that model is to keep the Python process alive or re-train from scratch. This is a significant gap for production workflows where models are trained in one environment(e.g. a batch job) and served or evaluated in another. Currently users are forced to reach for `pickle` or `joblib` manually, with no guarantee that the serialized object is compatible with the installed version of `causalML,` no standardized file format, and no integration with experiment tracking tools like `MLflow`

add `SerializableLearner` mixin that all base learners (BaseTLearner, BaseSLearner, BaseXLearner, BaseRLearner, BaseDRLearner) inherit from, exposing two methods:
```python
pythonlearner.save("model.causalml")
learner = BaseTRegressor.load("model.causalml")
```
every saved file would embed `__causalml_version__` and `__learner_class__` metadata. On load, if the installed version does not match, a `CausalMLVersionMismatchWarning` is raised so users are never silently running a stale model. `MLflow` integration would be added as an optional thin wrapper via `to_mlflow()` and `from_mlflow(),` raising a helpful `ImportError` if `MLflow` is not installed

using raw `joblib.dump` / `joblib.load` directly works but puts the burden on the user to handle versioning, class validation, and path conventions themselves. There is no way to know whether a saved file was produced by the same version of CausalML, which makes sharing models across teams or environments risky. A standardized mixin solves this once at the library level rather than in every user's codebase

**Additional context**

genrichez:
Closes #892.

## Summary

Adds `save()` and `load()` methods to all causal meta-learners (T, S, X, R, DR) via a `SerializableLearner` mixin on `BaseLearner`. Models are persisted using joblib and include metadata (causalml version, python version, class name, timestamp) to prevent stale model execution in production.

## Usage

```python
from causalml.inference.meta import BaseTRegressor, load_learner

# Train
learner = BaseTRegressor(learner=XGBRegressor())
learner.fit(X=X, treatment=treatment, y=y)

# Save
learner.save("model.causalml")

# Load (type-safe, checks class matches)
loaded = BaseTRegressor.load("model.causalml")

# Load (generic, no class check)
loaded = load_learner("model.causalml")
```

## Safety features

- **Version mismatch warning:** if the model was saved with a different causalml version, a `CausalMLVersionMismatchWarning` is raised on load
- **Class mismatch error:** loading a T-learner file as an S-learner raises `ValueError`
- **Unfitted guard:** calling `save()` before `fit()` raises `ValueError`
- **Missing file:** `FileNotFoundError` on bad path
- **Backwards compatible:** loading a raw joblib dump (no metadata wrapper) works with a warning

## Files changed

- `causalml/inference/meta/serialization.py` (new) - the `SerializableLearner` mixin
- `causalml/inference/meta/base.py` - added mixin to `BaseLearner` inheritance
- `causalml/inference/meta/__init__.py` - exported `load_learner` and `CausalMLVersionMismatchWarning`
- `tests/test_serialization.py` (new) - 19 unit tests

## Testing

- 19 unit tests covering round-trips for all learner types, safety checks, metadata validation, and edge cases
- Verified existing meta-learner tests pass without modification
- Tested with production scenarios: train-serve separation, champion-challenger deployment, model versioning for drift detection, compliance audit trail extraction, and checkpoint rollback

### DIFF
--- a/causalml/inference/meta/__init__.py
+++ b/causalml/inference/meta/__init__.py
@@ -10,3 +10,4 @@ from .xlearner import BaseXLearner, BaseXRegressor, BaseXClassifier
 from .rlearner import BaseRLearner, BaseRRegressor, BaseRClassifier, XGBRRegressor
 from .tmle import TMLELearner
 from .drlearner import BaseDRLearner, BaseDRRegressor, BaseDRClassifier, XGBDRRegressor
+from .serialization import load_learner, CausalMLVersionMismatchWarning
--- a/causalml/inference/meta/base.py
+++ b/causalml/inference/meta/base.py
@@ -7,6 +7,7 @@ from sklearn.base import BaseEstimator, clone
 from tqdm import tqdm
 
 from causalml.inference.meta.explainer import Explainer
+from causalml.inference.meta.serialization import SerializableLearner
 from causalml.inference.meta.utils import (
     check_p_conditions,
     filter_mask,
@@ -47,7 +48,7 @@ def _fit_bootstrap_clone(learner_template, X, treatment, y, p, seed, bootstrap_s
     return learner_b
 
 
-class BaseLearner(BaseEstimator, metaclass=ABCMeta):
+class BaseLearner(SerializableLearner, BaseEstimator, metaclass=ABCMeta):
     """Base class for all causalml meta-learners.
 
     Inheriting ``sklearn.base.BaseEstimator`` gives every subclass:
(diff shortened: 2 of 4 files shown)


jeongyoonlee (approved):
LGTM — thanks for the contribution, and welcome! Great first PR to causalml: the `SerializableLearner` mixin is clean, the metadata/version-mismatch guardrails match the #892 scope, and the test coverage (round-trip per learner, class/version/unfitted guards, bootstrap-ensemble, raw-joblib fallback) is thorough. Verified all 19 tests pass locally and the MRO plays nicely with the `BaseEstimator` refactor.

PR #930, opened 2026-07-04
Add DR and plug-in T scoring metrics for CATE model evaluation

### LINKED ISSUE #922: Add CATE model selection via surrogate metrics (R-score, DR-loss)
when using causalml in practice, i will often train S-learner, T-learner, X-learner, and DR-learner on the same data and have no principled way to pick between them. Standard cross validation does not work here because you never observe both potential outcomes for the same unit, so there's no ground truth to score against.

R-loss that powers the R-learner's training loop (rlearner.py lines 152, 495, 636) is exactly the right idea, but it's only used internally and never exposed as a standalone scoring function. `EconML` users already have this via `econml.score.RScorer` . Causalml users currently don't have an equivalent

add a `CATEModelSelector` to `causalml.metrics` that scores and compares fitted meta-learners using observed-data surrogate metrics, without needing access to true counterfactual outcomes...

metrics worth shipping in a first pass,
1. R-score (Nie and Wager, 2021): fits residual models for outcome and treatment, then measures how much CATE variance the model explains. Same R-loss already inside `BaseRLearner.fit()` , just exposed as a standalone function
2. DR pseudo-outcome loss (Kennedy 2023): constructs doubly-robust pseudo-outcomes and scores the CATE estimator against them. Mahajan et al. (2024) showed DR-based metrics dominate across 78 datasets
3. Plug-in T-score: uses a T-learner on held-out data as a CATE proxy. Mahajan et al. found this is never dominated across datasets despite being a simple baseline

rough API sketch
```python
from causalml.metrics import CATEModelSelector

selector = CATEModelSelector(
    models={'s': s_learner, 't': t_learner, 'x': x_learner, 'dr': dr_learner},
    metric='dr_loss',  # or 'r_score', 'plug_in_t'
    cv=5,
)
selector.fit(X, treatment, y)

selector.scores_       # {'s': 0.31, 't': 0.47, 'x': 0.61, 'dr': 0.58}
selector.best_model()  # 'x'
selector.plot_scores()
```

no new heavy dependencies needed. Nuisance models reuse causalml's existing propensity module and `sklearn`

using EconML's RScorer directly works but requires fitting nuisance components twice across two libraries with no integration between them.
picking by held-out outcome MSE is what most people default to, but outcome MSE doesn't measure CATE accuracy. A model can have low outcome MSE while getting the treatment effect completely wrong.

**Additional context**
happy to start with just the R-score as a standalone function in `causalml.metrics` and build the selector wrapper from there, if that's a better way to scope an initial PR.

aman-coder03:
## Proposed changes

this pr adds two new observed-data surrogate metrics for evaluating fitted cate models without access to true counterfactual outcomes

### what's included
- added `compute_dr_pseudo_outcomes()` to construct cross-fitted doubly robust (aipw) pseudo-outcomes that can be reused across different evaluation methods
- added `dr_score()` to evaluate cate models using mean squared error against the dr pseudo-outcomes
- added `plug_in_t_score()` to evaluate cate models against a cross-fitted plug-in t-learner cate proxy
- reused the same pseudo-outcomes with `rate_score()` to avoid fitting nuisance models multiple times
- added optional half-sample bootstrap confidence intervals and p-values, following the existing `rate_score()` implementation
- added unit tests covering pseudo-outcome generation, model ranking, bootstrap inference, error handling, and compatibility with `rate_score()`
- added documentation explaining the methodology and intended use of the new metrics

closes #922 

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/causalml/metrics/__init__.py
+++ b/causalml/metrics/__init__.py
@@ -26,6 +26,12 @@ from .visualize import (
 )  # noqa
 from .visualize import auuc_score, qini_score  # noqa
 from .rate import get_toc, rate_score, plot_toc  # noqa
+from .cate_scoring import (
+    compute_dr_pseudo_outcomes,
+    dr_score,
+    plug_in_t_score,
+)  # noqa
+from .sensitivity import Sensitivity, SensitivityPlaceboTreatment  # noqa
 from .sensitivity import Sensitivity, SensitivityPlaceboTreatment  # noqa
 from .sensitivity import (
     SensitivityRandomCause,
--- a/causalml/metrics/__init__.py
+++ b/causalml/metrics/__init__.py
@@ -32,7 +32,6 @@ from .cate_scoring import (
     plug_in_t_score,
 )  # noqa
 from .sensitivity import Sensitivity, SensitivityPlaceboTreatment  # noqa
-from .sensitivity import Sensitivity, SensitivityPlaceboTreatment  # noqa
 from .sensitivity import (
     SensitivityRandomCause,
     SensitivityRandomReplace,
--- a/causalml/metrics/cate_scoring.py
+++ b/causalml/metrics/cate_scoring.py
@@ -1,0 +1,473 @@
+from copy import deepcopy
+import logging
+
+import numpy as np
+import pandas as pd
+from scipy import stats
+from sklearn.model_selection import KFold
+
+from causalml.propensity import compute_propensity_score
+
+logger = logging.getLogger("causalml")
+
+
+def _resolve_outcome_learners(
+    learner, control_outcome_learner, treatment_outcome_learner
+):
+    """Resolve control/treatment outcome learners from constructor-style args.
+
+    Mirrors the ``learner`` / ``control_*_learner`` / ``treatment_*_learner``
+    convention used throughout ``causalml.inference.meta`` (e.g. ``BaseDRLearner``,
+    ``BaseTLearner``).
+
+    Args:
+        learner (model, optional): a model used for both control and treatment
+            outcome regressions if the group-specific learners are not given
+        control_outcome_learner (model, optional): a model to estimate outcomes
+            in the control group
+        treatment_outcome_learner (model, optional): a model to estimate outcomes
+            in the treatment group
+
+    Returns:
+        (tuple): the resolved (control_outcome_learner, treatment_outcome_learner)
+    """
+    if (learner is None) and (
+        (control_outcome_learner is None) or (treatment_outcome_learner is None)
+    ):
+        raise ValueError(
+            "Either `learner` or both `control_outcome_learner` and "
+            "`treatment_outcome_learner` must be specified."
+        )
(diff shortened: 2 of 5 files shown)


jeongyoonlee (changes requested):
The core AIPW/T-learner construction and the docs are solid, but one blocker before merge:

**Both `compute_dr_pseudo_outcomes` and `plug_in_t_score` can crash on imbalanced treatment — plain `KFold` doesn't stratify.**
Each fold fits per-arm outcome models on `X[train_idx][w_train == 0]` and `X[train_idx][w_train == 1]`. With `KFold(shuffle=True)` (not `StratifiedKFold`), a fold's *training* split can end up with zero units of one arm, so `.fit()` on the empty slice raises:

```
ValueError: Found array with 0 sample(s) (shape=(0, n_features)) while a minimum of 1 is required
```

Reproduced on rare-treatment data — which is exactly the observational setting these surrogate metrics target — at 3 treated / n=200 / 5 folds (crashes on 1 of 50 seeds):

```python
w = np.zeros(200, int); w[rng.choice(200, 3, replace=False)] = 1
compute_dr_pseudo_outcomes(X, w, y, learner=LinearRegression(), n_folds=5, random_state=15)
# ValueError: Found array with 0 sample(s) ...
```

Please switch both functions to `StratifiedKFold(n_splits=n_folds, shuffle=True, random_state=...)` and split on the treatment vector (`cv.split(X, treatment)`), so every fold retains both arms. That's the standard splitter for cross-fitting AIPW / T-learner nuisances.

Non-blocking notes to follow separately.

jeongyoonlee (commented):
Two non-blocking nits:

1. `p_value` column (`_score_against_pseudo_outcome`): `z_stat = point / se` tests H0: loss = 0, but the loss is an MSE that is ~never 0, so the p-value collapses to ≈0 for every model and conveys nothing. This is meaningful in `rate_score` (H0: RATE = 0 = no targeting value) but not for a loss — the informative test is a paired model-vs-model comparison. Suggest dropping the `p_value` column or replacing it with the paired test.

2. `metrics/__init__.py` has a duplicated `from .sensitivity import Sensitivity, SensitivityPlaceboTreatment  # noqa` line (a second copy of an already-present import).

jeongyoonlee (commented):
The `p_value` removal missed the docstrings — they still list a `p_value` column for `return_ci=True`. Please scrub the remaining references in `cate_scoring.py`:

- `_score_against_pseudo_outcome` — the `return_ci` arg ("…confidence intervals and p-values") and the Returns ("…CI bounds, and p-value (H0: loss = 0)…").
- `dr_score` — the `return_ci` arg and the Returns (same two).
- `plug_in_t_score` — the `return_ci` arg and the Returns (same two).

Also the `_score_against_pseudo_outcome` summary line still says it "mirrors the half-sample bootstrap CI/p-value pattern in `rate.py`" — drop the p-value half.

jeongyoonlee (approved):

PR #932, opened 2026-07-06
feat: extend serialization to CausalTree, CausalForest, UpliftTree, U…

### LINKED ISSUE #931: Extend save/load serialization to non-meta learners (causalml.inference.*)
### Summary

#929 (closes #892) added `save()`/`load()` + a generic `load_learner()` to the **meta-learners** via a `SerializableLearner` mixin on `causalml.inference.meta.base.BaseLearner`, with joblib persistence and version/class metadata guardrails. This issue tracks extending that capability to the rest of `causalml.inference.*` (uplift trees, causal trees/forest, IV, and the neural-net learners), and records the design decisions that extension requires.

### Two generalizations the mixin needs first

The mixin is generic in its useful parts (joblib+metadata wrapper, version/class checks, `load_learner`), but it's coupled to meta-learners in two places:

1. **Fitted-check is meta-specific.** `save()` guards with `hasattr(self, "t_groups")` (`serialization.py:99`). `t_groups` is a meta/DRIV concept — uplift trees signal fitted-ness with `fitted_uplift_tree` / `uplift_forest`, causal trees/forest with sklearn's `tree_` / `estimators_`. Replace with an overridable `_is_fitted()` hook (default to `sklearn.utils.validation.check_is_fitted(self)` for the sklearn-based estimators; override for uplift trees; keep `t_groups` for meta/DRIV).
2. **Location.** The mixin lives in `inference/meta/serialization.py` and is mixed into the *meta* base. To share it, lift it to a neutral module (e.g. `causalml/inference/serialization.py`) and mix it into each family's base class.

### Per-family feasibility (the real gate is picklability)

**Group A — joblib round-trips, low effort:**
- `CausalTreeRegressor`, `CausalRandomForestRegressor` — sklearn estimators, joblib works natively.
- `UpliftTreeClassifier`, `UpliftRandomForestClassifier` — plain Python; pickle-safe on Cython 3.x (verified during the #588 investigation).
- `IVRegressor`, `BaseDRIVLearner`/`BaseDRIVRegressor` — plain Python; DRIV already sets `self.t_groups` (`iv/drivlearner.py:116`), so it works with the mixin nearly unchanged.

Note: the sklearn-based and uplift learners are already joblib-dumpable today, so for them the mixin mainly adds the version/metadata guardrails and a consistent API — not new capability.

**Group B — neural nets, need framework-native serialization (do NOT use joblib):**
- `DragonNet` (TF and JAX) wraps a Keras/flax model — a joblib dump is fragile/broken; it already exposes native `save`/`load` (Keras H5 / flax).
- `CEVAE` wraps `torch` modules — needs `state_dict` via `torch.save`.
- Keep these on their native serialization; if unified under the shared API, the method should *delegate* to the framework-native path, not joblib.

### Naming / collision decision

`save()`/`load()` are **already defined** on the NN learners with a *different contract* than the mixin's:
- `tf/dragonnet.py:295,304` — `save(self, h5_filepath)`, `load(self, h5_filepath, ratio=1.0, dragonnet_loss=...)` (instance method, mutates `self`).
- `jax/dragonnet.py:484,507` — `save(self, path)`, `load(self, path, input_dim=None)`.

The mixin's `load()` is a **classmethod returning a new object**. Standardizing the cross-family verb on `save()`/`load()` would collide with these (and `load_model()` is a poor choice — "model" already denotes the *inner* base estimator, and it clashes with Keras' own `load_model`). 

**Decision:** standardize the shared cross-family verb on **`save_learner()` / `load_learner()`** — it matches causalml's own noun ("learner"), reuses the `load_learner()` name #929 already ships, and doesn't clobber the NN learners' native `save`/`load`. Meta-learners keep their current `save()`/`load()`/`load_learner()` surface (no collision there); no change needed to #929.

### Suggested increments
- **PR 1 (Group A):** lift + generalize the mixin (overridable `_is_fitted()`, neutral module) and apply to causal tree, causal forest, uplift tree, uplift forest, IV/DRIV. Round-trip tests per family.
- **PR 2 (Group B):** unify the API for DragonNet/CEVAE by delegating to Keras/flax/torch native serialization; guard tests with `importorskip`.

Refs: #892, #929.

genrichez:
Closes #931

## Summary

This PR extends the `SerializableLearner` mixin (introduced in #929 ) to Group A learners as outlined in the issue design spec.

### Changes

**New module: `causalml/inference/serialization.py`**

Lifted the mixin to a neutral location so it can be used by any learner family without circular imports. Added an overridable `_is_fitted()` hook that defaults to sklearn's `check_is_fitted` but can be customized per class.

**Group A learners now support `save()` / `load()`:**

| Learner | Fitted check |
|---------|-------------|
| CausalTreeRegressor | sklearn check_is_fitted (checks tree_) |
| CausalRandomForestRegressor | sklearn check_is_fitted (checks estimators_) |
| UpliftTreeClassifier | fitted_uplift_tree is not None |
| UpliftRandomForestClassifier | has uplift_forest attribute |
| BaseDRIVLearner (and subclasses) | has t_groups attribute |
| IVRegressor | has iv_fit attribute |

**Backward compatibility preserved:**

The old import path (`from causalml.inference.meta.serialization import ...`) still works via re-export.

**For Cython uplift classes:**

Since UpliftTreeClassifier and UpliftRandomForestClassifier are defined in a .pyx file, the mixin methods are injected at import time in `causalml/inference/tree/__init__.py`. This avoids modifying the Cython source while giving users the same API.

### Tests

21 new tests in `tests/test_serialization_extended.py`:
- Round-trip save/load for each learner family
- Unfitted model raises ValueError
- Class mismatch raises ValueError
- Generic `load_learner()` works across families
- Version mismatch produces warning (not error)
- Multiple save/load cycles
- Backward compat import path
- Cross-family type safety

All 40 serialization tests pass (19 original + 21 new).


jeongyoonlee (commented):
One gap: `BaseDRIVLearner` gets the mixin + `_is_fitted` and is listed as supported, but `test_serialization_extended.py` only covers CausalTree/Forest, UpliftTree/Forest, and IV — no DRIV. A `BaseDRIVLearner`/`DRIVLearner` save→load→predict round-trip would close it.

Everything else checks out — the round-trips, the `_is_fitted` hooks, and the import-time patching of the Cython uplift classes all look correct.

genrichez:
> One gap: `BaseDRIVLearner` gets the mixin + `_is_fitted` and is listed as supported, but `test_serialization_extended.py` only covers CausalTree/Forest, UpliftTree/Forest, and IV — no DRIV. A `BaseDRIVLearner`/`DRIVLearner` save→load→predict round-trip would close it.
> 
> Everything else checks out — the round-trips, the `_is_fitted` hooks, and the import-time patching of the Cython uplift classes all look correct.

Thanks for the review! Added the BaseDRIVLearner round-trip tests in dd93ac8 :  covers save→load→predict equality, unfitted save guard, generic load_learner dispatch, and metadata validation. All passing.

jeongyoonlee (approved):
Thanks for adding the `BaseDRIVLearner` round-trip tests — that closes the gap. Approving.

One optional nit for a follow-up (non-blocking): `tests/test_serialization_extended.py` hardcodes `RANDOM_SEED = 42`; the project convention is to import it from `tests/const.py` (as the cate-scoring tests do). `CONTROL_NAME = 0` is fine to keep local since the causal forest needs an integer label.

PR #933, opened 2026-07-07
Improve SensitivityMSM robustness and test coverage

aman-coder03:
## Proposed changes
follow ups on the review comments from #925 
This addresses several non-blocking follow-up suggestions from the review of the `SensitivityMSM` implementation.
changes focus on improving robustness, numerical stability, and test coverage without changing the public API or the underlying MSM estimation logic

Specifically,
- replaces the unsupported-learner blocklist with an explicit allowlist of supported learner families (S-, T-, and DR-learners), so unknown learner types fail safely instead of being treated as supported
- expands `test_SensitivityMSM` to cover S-, T-, and DR-learners instead of only the T-learner
- makes the MSM tests deterministic by seeding the synthetic data generation using `RANDOM_SEED`
- clips propensity bounds before division in the MSM bound computation to avoid numerical issues when propensity estimates are extremely close to 0 or 1

The documentation wording regarding the theoretical sharpness of the bound and the minor double-fit optimization are intentionally left for future work, as discussed in the previous review

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/causalml/metrics/sensitivity.py
+++ b/causalml/metrics/sensitivity.py
@@ -255,18 +255,19 @@ class Sensitivity:
 
         return summary_df
 
-    # Learners whose fit_predict(return_components=True) does NOT return
-    # potential-outcome regressions (mu0, mu1). X-learner returns two CATE
-    # estimates from its tau models instead (xlearner.py); R-learner has no
-    # outcome-regression decomposition at all. Both are rejected explicitly
-    # rather than silently misinterpreted.
-    _UNSUPPORTED_POTENTIAL_OUTCOME_LEARNERS = (
-        "BaseXLearner",
-        "BaseXRegressor",
-        "BaseXClassifier",
-        "BaseRLearner",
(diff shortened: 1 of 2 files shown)


jeongyoonlee (commented):
The allowlist matches `type(learner).__name__` against base-class names, so concrete S/T/DR subclasses are rejected: `XGBTRegressor`, `MLPTRegressor`, `LRSRegressor`, and `XGBDRRegressor` all hit the `NotImplementedError` here despite being supported. Use `isinstance(learner, (BaseSLearner, BaseTLearner, BaseDRLearner))` instead of name membership — it covers subclasses and still rejects X/R and unknown types.

The `test_SensitivityMSM` parametrization uses only the base classes (all in the allowlist), so this isn't caught — please add a concrete subclass (e.g. `XGBTRegressor`) to the parametrization.

jeongyoonlee (approved):
LGTM

PR #934, opened 2026-07-07
Fix NearestNeighborMatch crash when the control pool runs below `ratio`

arpitjain099:
## Summary

`NearestNeighborMatch.match()` with `replace=False` crashes on datasets where the pool of still-unmatched controls shrinks to `ratio` or fewer units. It calls `np.argpartition(dist, self.ratio)` on the remaining controls, and once that pool holds `ratio` or fewer elements argpartition raises `ValueError: kth(=...) out of bounds`. A balanced 1:1 dataset hits this on the very last treatment unit, so plain 1:1 matching without replacement fails on any balanced (or control-scarce) input.

The fix caps the partition index at the last available position and stops early once no unmatched controls remain. The abundant-control path is untouched, so existing behavior and the current tests are unchanged.

## Test plan

- Added a regression test in `tests/test_match.py` covering the balanced 1:1 case (previously raised `kth out of bounds`) and the control-scarce case.
- `pytest tests/test_match.py` passes locally.

### DIFF
--- a/causalml/match.py
+++ b/causalml/match.py
@@ -206,8 +206,17 @@ class NearestNeighborMatch:
                     match_to.loc[match_to.unmatched, score_col]
                     - match_from.loc[from_idx, score_col]
                 )
-                # Gets self.ratio lowest dists
-                to_np_idx_list = np.argpartition(dist, self.ratio)[: self.ratio]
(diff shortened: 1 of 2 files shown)


jeongyoonlee (approved):
LGTM

PR #935, opened 2026-07-08
Add `rlearner_score` (R-loss) for CATE model evaluation, EconML RScorer parity

aman-coder03:
## Proposed changes

2nd PR following #930 per @jeongyoonlee's scoping notes on #922  
Adds `rlearner_score()` as a standalone R-loss (Nie & Wager, 2021) scoring function for CATE model evaluation, giving EconML `RScorer` parity. also factors the cross-fitting logic out of `BaseRLearner.fit()` so the R-learner's own training loop and the new standalone scorer share the same nuisance-fitting code instead of duplicating it.

what's included,
- `compute_r_residuals()` in `causalml/propensity.py`, cross-fitted (stratified on treatment) outcome/treatment residuals `(y - m_hat(X), w - e_hat(X))`. lives in `propensity.py` rather than `inference.meta` or `metrics.cate_scoring` to avoid a real circular import between `causalml.metrics` and `causalml.inference.meta`, `propensity.py` is the only module with zero dependency on either. hit an actual `ImportError` tracing this down when it was first placed in `cate_scoring.py`
- `rlearner_score()` in `causalml/metrics/cate_scoring.py`, scores fitted CATE models via R-loss = `mean[(y_residual - w_residual * tau_hat(X))**2]`. named `rlearner_score` rather than `r_score` per review feedback on #922 (too close to `rate_score`). shares bootstrap CI logic with `dr_score()`/`plug_in_t_score()` via a new `_bootstrap_loss_ci()` helper. no `p_value` column, same reasoning as #930
- `BaseRLearner.fit()` refactored to call `compute_r_residuals()` instead of its own inline `cross_val_predict` block. scoped to `BaseRLearner.fit()` only, `BaseRClassifier.fit()` and `XGBRRegressor.fit()` keep their existing inline logic for now
- 10 new tests in `test_cate_scoring.py`, mirroring the existing DR/plug-in-T test shapes

worth flagging for review...
- switches `BaseRLearner.fit()`'s outcome-model cross-fitting from plain `KFold` to `StratifiedKFold` on treatment, for consistency with the propensity/DR nuisance splitting elsewhere (and the #930 rare-treatment fix). `model_mu` estimates `E[Y|X]` unconditionally so it was never exposed to the empty-arm crash itself, this is a consistency change, not a bugfix, but fold assignment for existing R-learner fits will differ run to run vs. before given the same `random_state`
- removed `self.cv = KFold(...)` from `BaseRLearner.fit()`, dead code once `compute_r_residuals()` builds its own splitter internally, and nothing else in the class reads `self.cv`

scoped out/follow-ups...
- `BaseRClassifier.fit()` and `XGBRRegressor.fit()` still use inline `cross_val_predict` rather than `compute_r_residuals()`
- `compute_r_residuals()` does `X = np.asarray(X)` internally, which is fine for `metrics.cate_scoring` callers but means `BaseRLearner.fit()` now coerces `X` to numpy for this one step, diverging from the polars/pandas-native contract the rest of `fit()` follows (`collect_if_lazy`, `filter_mask`, etc.). not fixed here since it'd need either a native-X rewrite of `compute_r_residuals()` or branching logic, flagging rather than blocking on it
- `docs/methodology.rst` and `docs/refs.bib` updates (extending the "CATE Model Evaluation" section from #930, adding a citation for the R-learner paper) aren't in this PR yet, will add here or as a doc-only follow-up

closes the remaining scope from #922 

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/causalml/inference/meta/rlearner.py
+++ b/causalml/inference/meta/rlearner.py
@@ -150,6 +150,7 @@ class BaseRLearner(BaseLearner):
             n_folds=self.n_fold,
             random_state=self.random_state,
             n_jobs=self.cv_n_jobs,
+            compute_w_residual=False,
         )
         yhat = y_np - y_residual
         # Fit the nuisance outcome model on the full data so it can be
--- a/causalml/inference/meta/rlearner.py
+++ b/causalml/inference/meta/rlearner.py
@@ -16,7 +16,7 @@ from causalml.inference.meta.utils import (
     get_xgboost_objective_metric,
     get_weighted_variance,
 )
-from causalml.propensity import ElasticNetPropensityModel
+from causalml.propensity import ElasticNetPropensityModel, compute_r_residuals
 
 logger = logging.getLogger("causalml")
 
@@ -135,10 +135,6 @@ class BaseRLearner(BaseLearner):
             else deepcopy(self.learner)
         )
         self.model_p = self.propensity_learner
-        # Build CV splitter from stored n_fold / random_state.
-        self.cv = KFold(
-            n_splits=self.n_fold, shuffle=True, random_state=self.random_state
-        )
 
         self.models_tau = {group: deepcopy(self.model_tau) for group in self.t_groups}
         self.vars_c = {}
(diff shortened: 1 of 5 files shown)


jeongyoonlee (changes requested):
`BaseRLearner.fit()` routes the outcome cross-fit through `compute_r_residuals()` but drops `w_residual` (`_`) and passes neither `p` nor `propensity_learner` — so `compute_r_residuals` cross-fits a full propensity model (default ElasticNet, per fold) on every fit and discards it. `fit()` already computed `p` above; this is wasted work: `fit()` is ~4× slower even when the caller supplies `p` (0.8s → 3.3s at n=8k, p=20). Add an outcome-only path to `compute_r_residuals` (skip propensity when `w_residual` isn't needed) and call that here.

jeongyoonlee:
Merging master left a duplicate in `fit()`: `self.model_p = self.propensity_learner` is now set at both line 122 (from the merged #937) and line 143. Please drop line 143.

jeongyoonlee (approved):
LGTM

Issue #571, opened 2022-11-13
Why does using  UpliftRandomForestClassifier  consume so much memory？


longweiwei:
**Describe the bug**

Train the model using UpliftRandomForestClassifier and using default configuration, When the amount of training data reaches 4.5G（about 2.9 million rows of data, 440 features）, the required memory exceeds 80G. I'm not sure if this is a bug,

**To Reproduce**
Can be reproduced the situation using random data。
```
from causalml.inference.tree import UpliftRandomForestClassifier
from causalml.metrics import plot_gain
from sklearn.model_selection import train_test_split
import causalml
causalml.__version__

from causalml.inference.tree import UpliftTreeClassifier
import numpy as np
from sys import getsizeof
import pandas as pd

treatment = np.random.randint(0, 2, (2900000)).astype(np.float32)
y =  np.random.randint(0, 2, (2900000)).astype(np.float32)
feats = np.random.randn(2900000, 440).astype(np.float32)
treatment_bak = ['treat' if v==1 else 'control' for v in treatment.tolist()]

uplift_model = UpliftRandomForestClassifier(control_name='control')
uplift_model.fit(X=feats,
                 treatment=np.array(treatment_bak),
                 y=y)
```
**Expected behavior**
Is there any way to reduce memory usage besides reducing the amount of data?

**Screenshots**

**Environment (please complete the following information):**
 - OS: ubuntu
 - Python Version: 3.6,
 - Versions of Major Dependencies (`pandas`, `scikit-learn`, `cython`):  pandas== 1.0.4 , scikit-learn==0.24.2, cython==0 0.29.24 

**Additional context**

longweiwei:
@Pawka @mway @darthtrevino @shannonmoeller 
hi! can you help with the issue

zyq11223:
because the data type of the treatment variable is str object , it cost too much memory, and the library does not support int type

jbbqqf:
Hi! Triaging older issues — I think this one can be closed unless someone has a fresh reproducer that isolates a real leak.

[...]

PR #936, opened 2026-07-08
Lazy-fit `model_mu` in R-learner so it's only trained when `return_components` is used

aman-coder03:
## Proposed changes
follow up to #923  `model_mu` (the nuisance outcome model) was being fit on the full training data inside `fit()` every time, even when nobody ever calls `predict(..., return_components=True)` ,that's wasted work for the common case

this defers the fit: `fit()` now just caches the training data, and `model_mu` only gets fit the first time `predict(return_components=True)` is actually called, then cached so it's not re-fit on later calls

also fixes a related bug this uncovered: the bootstrap loops in `fit_predict(return_ci=True)` and `estimate_ate(bootstrap_ci=True)` save/restore `model_mu` and `models_tau` around resampling, but weren't saving/restoring the new cached training data used for the lazy fit. without the fix, calling `predict(return_components=True)` after a bootstrap run would silently fit `model_mu` on stale bootstrap-resampled data instead of the real training set

applies to `BaseRLearner`, `BaseRClassifier`, and `XGBRRegressor`

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/causalml/inference/meta/rlearner.py
+++ b/causalml/inference/meta/rlearner.py
@@ -149,9 +149,11 @@ class BaseRLearner(BaseLearner):
         yhat = cross_val_predict(
             self.model_mu, X, y_np, cv=self.cv, n_jobs=self.cv_n_jobs
         )
-        # Fit the nuisance outcome model on the full data so it can be
-        # reused by predict(return_components=True).
-        self.model_mu.fit(X, y_np)
+        # Defer fitting the nuisance outcome model on the full data until
+        # predict(return_components=True) actually needs it.
+        self._model_mu_fitted = False
+        self._mu_fit_X = X
+        self._mu_fit_y = y_np
 
         for group in self.t_groups:
             mask = (treatment_np == group) | (treatment_np == self.control_name)
@@ -215,6 +217,10 @@ class BaseRLearner(BaseLearner):
 
         if not return_components:
             return te
+
+        if not self._model_mu_fitted:
+            self.model_mu.fit(self._mu_fit_X, self._mu_fit_y)
+            self._model_mu_fitted = True
 
         if p is None:
             if not hasattr(self, "propensity_model"):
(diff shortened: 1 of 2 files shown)


jeongyoonlee (changes requested):
Deferring the `model_mu` fit requires pinning the full training set on the instance (`self._mu_fit_X`, `self._mu_fit_y`), which inflates a fitted R-learner from 2.7 KiB to 4.8 MiB at n=20k/p=30 (pickle). With `save()`/`load()` added in #929/#932, every saved R-learner now carries its training data — a steep cost to avoid one `model_mu.fit(X, y)`, which is cheap for the default learners. Either keep the eager fit, or exclude the cached X/y from serialization so on-disk size doesn't regress.

The new tests cover only `BaseRLearner`; `BaseRClassifier` and `XGBRRegressor` get the same lazy machinery but no test — please add at least the classifier path.

Note this overlaps #935, which also rewrites `BaseRLearner.fit()` (and keeps the eager `model_mu` fit this PR removes) — they'll need to be sequenced.

jeongyoonlee:
Recommend closing this. The pickle blowup is resolved by `__getstate__`, but the design now permanently retains the full training `X`/`y` on every fitted R-learner — measured `(20000, 30)` held for the object's lifetime, and not cleared even after the lazy fit fires. For this PR's own motivating case (`return_components` never used), that replaces a KiB-sized fitted `model_mu` with the entire training set in RAM, unbounded in dataset size — which cuts against the memory-at-scale concerns in #571.

The upside is skipping one `model_mu.fit()` — on top of the `n_fold` fits the cross-fit already does (~24% of `fit()` for an expensive outcome learner, marginal otherwise). Not worth permanent full-dataset retention. It also conflicts with #935 on the same `fit()` lines. Suggest keeping the eager fit.

aman-coder03:
agreed, closing this PR
the permanent full-dataset retention isn't worth avoiding one `model_mu.fit()` call, especially since it works against #571 memory concerns rather than for them
reverting to eager fit is the right call here and #935 should proceed without this conflict

PR #937, opened 2026-07-08
Use custom propensity_learner on R-learner's first fit

arpitjain099:
## Proposed changes

The R-learner ignores a user-supplied `propensity_learner` on the first `fit()` of a fresh object. In each `fit()` override the line `self.model_p = self.propensity_learner` runs *after* `_set_propensity_models()`, but `_set_propensity_models` reads `self.model_p` (via `hasattr`) to decide whether to use the custom propensity model or fall back to the default `ElasticNetPropensityModel`. On a new object `model_p` doesn't exist yet, so the first fit quietly uses the default and the custom learner is dropped.

Because the attribute persists on the object, a second `fit()` of the same instance *does* pick up the custom learner, so the same call gives different nuisance estimates depending on fit order. The fix moves the `model_p` assignment ahead of the propensity-model fitting in all three affected paths (`BaseRLearner`, `BaseRClassifier`, `XGBRRegressor`).

## Types of changes

## Test plan

- Added `test_BaseRLearner_uses_custom_propensity_learner_on_first_fit` in `tests/test_meta_learners.py`: it fits a fresh R-learner with a distinctive `ElasticNetPropensityModel` subclass and asserts the trained `propensity_model` is that subclass. It fails on `master` (default ElasticNet is used) and passes with this change.
- The R-learner test subset (`-k "RLearner or RRegressor or RClassifier"`, 21 tests) passes locally.

## Checklist

### DIFF
--- a/causalml/inference/meta/rlearner.py
+++ b/causalml/inference/meta/rlearner.py
@@ -115,6 +115,12 @@ class BaseRLearner(BaseLearner):
         self.t_groups = np.unique(treatment_np[treatment_np != self.control_name])
         self.t_groups.sort()
 
+        # Resolve model_p before fitting propensity models so a user-supplied
+        # propensity_learner is honored on the first fit. _set_propensity_models
+        # reads self.model_p, so setting it afterwards meant the first fit of a
+        # fresh object silently fell back to the default ElasticNet.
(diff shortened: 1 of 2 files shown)


jeongyoonlee (approved):
LGTM

PR #940, opened 2026-07-14
Expose matched index pairs as NearestNeighborMatch.matched_indexes_ (#621)

### LINKED ISSUE #621: Add matched indexes as NearestNeighborMatch class attribute

jeongyoonlee:
## Proposed changes

Expose the matched index pairs computed inside `NearestNeighborMatch.match()` as a fitted attribute `matched_indexes_` — a two-column `pandas.DataFrame` (`from`, `to`) where each row is one matched pair. Useful for joining matched pairs back to upstream metadata or auditing the matching outcome without re-running the algorithm.

Previously these `(from, to)` pairs were computed and then discarded; only the joined DataFrame was returned. Under `ratio > 1` a single `from` index can appear multiple times against distinct `to` indices, so the attribute captures the pair mapping (not just the deduplicated from-set). Both the replacement (`NearestNeighbors`) and no-replacement (caliper-loop) paths are covered. `match_by_group` calls `match()` per group, so the attribute reflects the most recent call.

Fixes #621.

## Credit

Reimplements the work of **@jbbqqf** in #897, which is CLA-blocked and cannot be merged directly. Full credit to @jbbqqf for the original implementation and analysis; this PR re-lands the same feature against current `master` under a CLA-clean author.

## Types of changes

## Tests

- New `test_nearest_neighbor_match_exposes_matched_indexes` in `tests/test_match.py` covering both branches (schema, consistency with the returned matched frame, and pair-level granularity under `replace=True, ratio=2`). Fails on `master` with `AttributeError`, passes here.
- Full `tests/test_match.py` — 6 passed.

### DIFF
--- a/causalml/match.py
+++ b/causalml/match.py
@@ -97,6 +97,19 @@ class NearestNeighborMatch:
             seed
         n_jobs (int): The number of parallel jobs to run for neighbors search.
             None means 1 unless in a joblib.parallel_backend context. -1 means using all processors
+
+    Fitted attributes (populated after :meth:`match`):
+        matched_indexes_ (pandas.DataFrame): two-column dataframe with the
+            ``(from, to)`` pairs of original data indices produced by the most
+            recent :meth:`match` call. ``from`` corresponds to the matching
+            source group (treatment if ``treatment_to_control`` else control);
+            ``to`` is the matched counterpart from the opposite group. Each row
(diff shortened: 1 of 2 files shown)


ras44 (approved):
LGTM

PR #941, opened 2026-07-14
Validate binary outcome in filter_LR / filter_D (#349)

### LINKED ISSUE #349: Clarify the limitations (specifically the label) of the current implementation of `filter methods`

jeongyoonlee:
## Proposed changes

The likelihood-ratio filter (`filter_LR`) and the divergence-based filters (`filter_D` — `KL` / `ED` / `Chi`) model a **binary** conversion outcome. Passing a continuous or multi-class outcome previously produced meaningless statistics **silently**, with no error.

This adds a module-level `_check_binary_outcome()` helper, called at the top of `filter_LR` and `filter_D`, that raises a clear `ValueError` naming the offending outcome column and pointing to `filter_F` (the OLS F-test filter, which tolerates continuous outcomes). The module docstring now documents the binary-only constraint. `filter_F` is unchanged.

Fixes #349.

## Credit

Reimplements the work of **@jbbqqf** in #898, which is CLA-blocked and cannot be merged directly. Full credit to @jbbqqf for the original implementation and analysis; this PR re-lands the same feature against current `master` under a CLA-clean author.

## Types of changes

## Tests

- New parametrized `test_filter_rejects_non_binary_outcome` over `["LR", "KL", "ED", "Chi"]` (4-class outcome → `ValueError` matching "binary") and `test_filter_f_accepts_continuous_outcome` (continuous outcome still works through `filter_F`). The rejects tests fail on `master` (no error raised), pass here.
- Full `tests/test_feature_selection.py` — 8 passed.

### DIFF
--- a/causalml/feature_selection/filters.py
+++ b/causalml/feature_selection/filters.py
@@ -1,7 +1,11 @@
-"""
-Filter feature selection methods for uplift modeling
+"""Filter feature selection methods for uplift modeling.
 
-- Currently only for classification problem: the outcome variable of uplift model is binary.
+.. note::
+    The likelihood-ratio filter (:func:`FilterSelect.filter_LR`) and the
+    divergence-based filters (:func:`FilterSelect.filter_D` -- ``KL``, ``ED``,
+    ``Chi``) require a **binary** outcome (values in ``{0, 1}``) and raise
+    ``ValueError`` otherwise. Use the F-test filter
+    (:func:`FilterSelect.filter_F`) for a continuous outcome.
 """
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


ras44 (approved):
👍

ras44 (approved):
LGTM

PR #943, opened 2026-07-15
Stop logloss from clipping the caller's prediction array in place

arpitjain099:
## Proposed changes

No filed issue for this one. I was reading through `causalml/metrics` and noticed `logloss` bounds `p` with `p[p < EPS] = EPS` and `p[p > 1 - EPS] = 1 - EPS`, which write straight into whatever array the caller handed in. So calling the exported metric edits your predictions as a side effect wherever a value sits at or past the bounds:

```python
>>> p = np.array([0.0, 1.0, 0.2, 0.8, 0.9])
>>> logloss(np.array([0, 1, 0, 1, 1]), p)
>>> p
array([1.e-15, 1.e+00, 2.e-01, 8.e-01, 9.e-01])
```

That bites if you reuse `p` for another metric afterwards, or just look at it. `classification_metrics` happens to pass `p[w]`, which is already a copy, so the internal path never showed it.

Switched to `np.clip`, which returns a new array and keeps the same bounds. I checked old vs new over 200 random inputs including 0.0/1.0 values, and the clipped arrays and the resulting `log_loss` came out identical, so this drops the side effect and nothing else.

## Types of changes

## Checklist

## Further comments

Two tests in `tests/test_metrics.py`. `test_logloss_does_not_modify_prediction` fails on master and passes here. `test_logloss_clips_degenerate_predictions` passes either way, it's just there so the clipping doesn't get dropped later.

`black` is clean.

### DIFF
--- a/causalml/metrics/classification.py
+++ b/causalml/metrics/classification.py
@@ -1,4 +1,5 @@
 import logging
+import numpy as np
 from sklearn.metrics import log_loss, roc_auc_score
 
 from .const import EPS
@@ -16,8 +17,7 @@ def logloss(y, p):
         bounded log loss error
     """
 
-    p[p < EPS] = EPS
-    p[p > 1 - EPS] = 1 - EPS
+    p = np.clip(p, EPS, 1 - EPS)
     return log_loss(y, p)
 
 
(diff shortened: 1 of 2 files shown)


jeongyoonlee (approved):
LGTM. Thanks for the contribution!

jeongyoonlee (approved):

PR #944, opened 2026-07-16
Widen the propensity model's cross-validated C grid

arpitjain099:
## Proposed changes

No filed issue, I was reading through `causalml/propensity` and the `Cs` value on `LogisticRegressionPropensityModel` looked off:

```python
"Cs": np.logspace(1e-3, 1 - 1e-3, 4),
```

That expands to about `[1.00, 2.16, 4.64, 9.98]`, so every candidate is `C >= 1`. `LogisticRegressionCV` searches `C` (the inverse of regularization strength), so this grid only ever lets it choose weak regularization and it can never reach `C < 1`. The whole search sits inside a single order of magnitude on the weak side.

It reads like the `1e-3` / `1 - 1e-3` bounds got carried over from the `l1_ratios` line right below it (where a `[0, 1]` range is exactly right) and from `clip_bounds`. For `Cs` those numbers land in the exponent of `logspace`, which is what collapses the range.

I changed it to `Cs=4`, which hands the job back to `LogisticRegressionCV`: it builds its own log grid between `1e-4` and `1e4` and keeps four points. Fitted models now search across both strong and weak regularization. This does shift the chosen penalty on existing pipelines, but toward the range the CV was meant to cover.

## Types of changes

## Checklist

## Further comments

Added `test_logistic_regression_propensity_model_cs_grid` in `tests/test_propensity.py`. It fits the model and asserts the resulting `model.Cs_` spans `min < 1 < max`. On master the grid is `[1.00, 2.16, 4.64, 9.98]` so it fails; with this change `Cs_` is `[1e-4, 4.6e-2, 21.5, 1e4]` and it passes. `black` is clean.

Happy to use an explicit grid like `np.logspace(-4, 4, 4)` instead if you'd rather keep it spelled out, or bump the point count.

### DIFF
--- a/causalml/propensity.py
+++ b/causalml/propensity.py
@@ -97,7 +97,7 @@ class LogisticRegressionPropensityModel(PropensityModel):
         kwargs = {
             "penalty": "elasticnet",
             "solver": "saga",
-            "Cs": np.logspace(1e-3, 1 - 1e-3, 4),
+            "Cs": 4,
             "l1_ratios": np.linspace(1e-3, 1 - 1e-3, 4),
             "cv": StratifiedKFold(
                 n_splits=(
--- a/tests/test_propensity.py
+++ b/tests/test_propensity.py
@@ -1,3 +1,5 @@
+import numpy as np
+
 from causalml.propensity import (
     ElasticNetPropensityModel,
     GradientBoostedPropensityModel,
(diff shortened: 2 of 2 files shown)


jeongyoonlee (approved):
LGTM. Thanks!

PR #958, opened 2026-07-17
Fix ATE column filled with NaN in SensitivitySelectionBias.summary

arpitjain099:
`SensitivitySelectionBias.summary()` builds the `ATE` column from a filtered slice:

    sensitivity_summary["ATE"] = sensitivity_summary[
        sensitivity_summary.alpha == 0
    ]["New ATE"]

The right side is a Series holding only the `alpha == 0` row, and it keeps that row's index label. Assigning it back to the full frame aligns on the index, so only the `alpha == 0` row keeps a value and every other row ends up NaN. The `ATE` column is meant to be the baseline (unconfounded) ATE shown next to `New ATE` at each confounding level, so it should be that one value on every row.

Pulling the scalar out and letting pandas broadcast it fixes the summary:

    sensitivity_summary["ATE"] = sensitivity_summary.loc[
        sensitivity_summary.alpha == 0, "New ATE"
    ].values[0]

Before, on a small synthetic run, 10 of 11 rows in the ATE column were NaN. After, the alpha=0 estimate shows on every row, which is what you want when comparing it against New ATE across the alpha range.

Existing SelectionBias/PlaceboTreatment/RandomCause sensitivity tests still pass. Glad to add a small regression check on the summary output if that's useful.

### DIFF
--- a/causalml/metrics/sensitivity.py
+++ b/causalml/metrics/sensitivity.py
@@ -565,9 +565,9 @@ class SensitivitySelectionBias(Sensitivity):
         sensitivity_summary["Method"] = sensitivity_summary[
             "Method"
         ] + sensitivity_summary["rsqs"].round(5).astype(str)
-        sensitivity_summary["ATE"] = sensitivity_summary[
-            sensitivity_summary.alpha == 0
-        ]["New ATE"]
+        sensitivity_summary["ATE"] = sensitivity_summary.loc[
+            sensitivity_summary.alpha == 0, "New ATE"
+        ].values[0]
         return sensitivity_summary[SUMMARY_COLS]
 
     @staticmethod
--- a/tests/test_sensitivity.py
+++ b/tests/test_sensitivity.py
@@ -228,6 +228,50 @@ def test_SensitivitySelectionBias():
     sens.plot(lls_bias_alignment, ci=True)
 
 
+def test_SensitivitySelectionBias_summary_ate():
+    # summary() used to assign the alpha == 0 "New ATE" as a Series, so pandas
+    # aligned it on index and every row other than alpha == 0 came back NaN.
+    np.random.seed(RANDOM_SEED)
+
+    y, X, treatment, tau, b, e = synthetic_data(
+        mode=1, n=100000, p=NUM_FEATURES, sigma=1.0
+    )
+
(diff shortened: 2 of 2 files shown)


jeongyoonlee:
Can you add the regression test you offered? It should assert SensitivitySelectionBias.summary()["ATE"] has no NaN and equals the alpha == 0 New ATE on every row. That's what would have caught the original index-alignment bug. Use RANDOM_SEED / the shared column constants from tests/const.py.

arpitjain099:
@jeongyoonlee 
Added in 17f1d07, `test_SensitivitySelectionBias_summary_ate`. It seeds with `RANDOM_SEED` and uses `TREATMENT_COL` / `SCORE_COL` / `OUTCOME_COL` / `NUM_FEATURES` from `tests/const.py`, and mirrors the data setup in the existing `test_SensitivitySelectionBias` so it exercises the same path.

It asserts both things you asked for: `summary()["ATE"]` has no NaN, and every row equals the `alpha == 0` New ATE.

One wrinkle worth mentioning. `summary()` returns `SUMMARY_COLS`, which doesn't include `alpha`, so the test can't just filter on it. Calling `causalsens()` again to get it would refit and could drift, so instead it recovers the alpha from the `Method` string, which already carries it:

```python
alphas = summary["Method"].str.extract(r"alpha@(-?[\d.eE+-]+),", expand=False).astype(float)
baseline = summary.loc[alphas == 0, "New ATE"]
assert len(baseline) == 1
assert np.allclose(summary["ATE"], baseline.values[0])
```

If you'd rather not depend on the Method text, the alternative is adding `alpha` to the returned frame, but that changes the public shape of `summary()` so I left it alone.

On verification, being straight with you: I couldn't run the suite locally, the Cython extensions don't build in my environment, so CI is the real check here. What I did verify directly is that the assertions actually discriminate. Reproducing the old and new assignment on a small frame with an `alpha == 0` row:

```
buggy  ATE column: [nan, nan, 0.6, nan, nan]   -> no-NaN False, all-equal False
fixed  ATE column: [0.6, 0.6, 0.6, 0.6, 0.6]   -> no-NaN True,  all-equal True
```

So the test fails on the pre-fix behavior and passes on the fixed one, which is the property you wanted it to have.

jeongyoonlee (approved):

Issue #946, opened 2026-07-17
[Tree unification 1/9] Criterion scaffold + kernel-backed single Uplift tree (KL/ED/Chi), parity vs legacy


jeongyoonlee:
Replace uplift's own pure-Python `DecisionTree` node class and recursive builder with the shared scikit-learn-vendored Cython kernel (`causalml/inference/tree/_tree/`) for the three simplest splitting criteria — **KL**, **ED (Euclidean)**, and **Chi-squared** — and prove numerical parity against the legacy `UpliftTreeClassifier`.

This is the foundation of the epic: it stands up the new parallel package and the criterion↔kernel plumbing. Regularization, normalization, honesty, pruning, and the forest are all **deferred** to later issues.

## Scope

- New package `causalml/inference/tree/_uplift/` (leading underscore — a sibling `uplift/` cannot coexist with `uplift.pyx`; the public surface is routed through `tree/__init__.py`, so the internal name is invisible).
- `_uplift/_criterion.pyx` + `.pxd` (**C++**, `# distutils: language = c++`): `UpliftClassificationCriterion` reusing the causal `NodeState` / `NodeSplitState` group-stats machinery (cimport/subclass from `..causal._criterion`, or lift the shared machinery into a common base if cleaner). Concrete `KLCriterion`, `EDCriterion`, `ChiCriterion` implementing `node_impurity` + `children_impurity` from the legacy `evaluate_KL/ED/Chi` formulas (`uplift.pyx` ~1032–1186), using `outcome_mean(g)` = P(Y=1|T=g).
- `_uplift/_tree.py`: `BaseUpliftDecisionTree(BaseDecisionTree)` + `UPLIFT_TREE_CRITERIA` string→class map + a `fit()` copying the causal `_tree.py` wiring but using the **stock** kernel splitters (`DENSE/SPARSE_SPLITTERS`) and **stock** `DepthFirst/BestFirstTreeBuilder` (no uplift regularization yet).
- Data prep: `(n_samples, n_groups)` NaN-masked **binary**-y matrix, control at col 0 (mirror `causaltree.py::_prepare_data`; coerce `y = (y > 0)` per `uplift.pyx`).
- Experimental estimator `_KernelUpliftTreeClassifier` — **not** wired into public `tree/__init__.py`; legacy stays the default.
- `setup.py`: register the new C++ extension(s).

## Verify

[...]

Issue #945, opened 2026-07-17
Unify the tree base — port Uplift trees onto the shared `_tree/` Cython kernel (v1.0 M1)


jeongyoonlee:
## Summary

CausalML has **two tree families with completely separate implementations**:

- **Causal trees** (`CausalTreeRegressor` / `CausalRandomForestRegressor`) already ride the shared scikit-learn-1.5.2-vendored Cython kernel in `causalml/inference/tree/_tree/` (`Tree`, `TreeBuilder`, `Splitter`, `RegressionCriterion`). The causal layer is a thin set of subclasses: a criterion (`causal/_criterion.pyx`), a builder (`causal/_builder.pyx`), and an estimator stack `BaseCausalDecisionTree(BaseDecisionTree)` → `CausalTreeRegressor` → `CausalRandomForestRegressor(ForestRegressor)`. Treatment is encoded in `y` as a `(n_samples, n_groups)` NaN-masked matrix, control at column 0.
- **Uplift trees** (`UpliftTreeClassifier` / `UpliftRandomForestClassifier`) do **not**. They live in a single **2,742-line** `causalml/inference/tree/uplift.pyx` with their own pure-Python `DecisionTree` node class, their own recursive builder (`growDecisionTreeFrom`), their own joblib bagging forest, their own 8 splitting criteria, and their own NaN handling. They compile to **C** (the kernel is **C++**), don't inherit `BaseEstimator`, and get `save`/`load`/`_is_fitted` **monkey-patched** onto them at import time in `tree/__init__.py`.

The **v1.0 roadmap RFC** ([Discussion #938](https://github.com/uber/causalml/discussions/938)) lists this unification under **M1 — Foundation & stability**. It is a **prerequisite** for two other M1 bullets — `check_estimator` conformance and serialization hygiene — because the uplift classes' non-`BaseEstimator` status blocks both. It also collapses the M2 profiling surface from two Cython tree engines to one, and deleting it after 1.0 GA would be a breaking-change hazard.

**Goal:** re-implement the uplift trees on top of the shared `_tree/` kernel — reusing its `Tree`, splitter, and builder scaffolding and the causal criterion's per-group state machine — while **preserving public class names, argument names, defaults, and prediction outputs**. Delivered as this epic broken into bite-size issues, executed in a **new parallel package** that proves numerical parity against the legacy code before a final switchover deletes the ~2.3 KLoC duplicate.

## Key enabling insight

[...]

PR #961, opened 2026-07-23
Add uplift split regularization to the kernel tree (n_reg/min_samples_treatment) (#947)

### LINKED ISSUE #947: [Tree unification 2/9] Uplift split regularization via a custom builder
Add the uplift-specific split regularization that the legacy tree applies during growth, via a custom builder that subclasses the shared kernel `TreeBuilder` (mirroring `causal/_builder.pyx`).

Depends on **Issue 1** (kernel-backed KL/ED/Chi tree scaffold).

## Scope

- `_uplift/_builder.pyx` + `.pxd`: `DepthFirst/BestFirstUpliftTreeBuilder(TreeBuilder)` mirroring `causal/_builder.pyx` — downcast the criterion to read per-group counts and:
  - enforce `min_samples_treatment` (minimum units per treatment group in each child),
  - apply the `n_reg` parent-shrinkage the legacy gain applies via `parentNodeSummary`,
  - `early_stopping_eval_diff_scale` (validation-set early stopping may fold in here, or split to its own issue if it doesn't fit the builder cleanly).
- Wire the estimator to select the uplift builder instead of the stock kernel builder.

## Verify

- `min_samples_treatment` respected (no child with fewer than the threshold per group).
- Parity vs legacy **defaults** (`n_reg=100`, `min_samples_treatment=10`) for KL/ED/Chi on a fixed seed.

## Reusable templates

- Builder subclass: `causalml/inference/tree/causal/_builder.pyx` + `.pxd` (`DepthFirst/BestFirstCausalTreeBuilder` add `min_group_samples` by downcasting the criterion).
- Legacy regularization: `uplift.pyx` `growDecisionTreeFrom` (~1968–2394) and the `n_reg` shrinkage in the gain.

Part of #945.

jeongyoonlee:
## Summary

Second step of the tree-unification epic (#945, follows #946/#959). Ports the
Rzepakowski et al. (2012) **parent-shrinkage regularization** onto the
kernel-backed uplift tree so it reproduces the legacy `UpliftTreeClassifier`
with regularization **on** — the prior PR only proved parity with it off.

Two knobs, matching the public `UpliftTreeClassifier`:

- **`n_reg`** (default 100): each node's `P(Y=1|T=g)` shrinks toward its parent's
  regularized probability with weight `n_reg` (in sample-size units) when
  evaluating a split's gain.
- **`min_samples_treatment`** (default 10): the shrinkage threshold *and* a hard
  admissibility floor — a candidate split whose left/right child has any group
  below it is rejected.

Still internal/experimental (`_KernelUpliftTreeClassifier`); the public default
remains the legacy tree.

## How it works

- **`_uplift/_criterion.pyx`** — the split gain
  `p·D_left + (1−p)·D_right − D_node` is now computed on *regularized* per-group
  probabilities: the node shrinks toward its parent, the children shrink toward
  the node (exact port of legacy `tree_node_summary` + the KL/ED/Chi
  `evaluate_*`). A split with any child group `< min_samples_treatment` gets gain
  `−inf`, so the search never selects it — the kernel analogue of the legacy
  `node_mst < min_samples_treatment → continue`. **Leaf predictions
  (`node_value`) stay raw** `n_pos/n`, matching legacy
  `uplift_classification_results` (regularization affects only growth, not the
  leaf estimate).

- **`_uplift/_builder.pyx`** (new) — `DepthFirst`/`BestFirst` builders that thread
  each node's regularized summary down to its children. The summary can't live in
  `tree.value` (which must stay raw for prediction), so the builder keeps a
  `node_id → regularized summary` map, setting it on the criterion before each
  `node_split` and reading it back after `node_value`. They also omit the stock
  `impurity ≤ EPSILON` early-leaf, which is incompatible with uplift growth
  (driven by the split gain; shrinkage further shrinks child divergences toward
  zero, which would otherwise stop growth prematurely and diverge from legacy).

- **`_tree.py` / `uplifttree.py`** — wire the new builders and expose `n_reg` /
  `min_samples_treatment`.

## Testing

`pytest tests/test_uplift_trees_kernel.py` — 29 passed.

- **New**: exact whole-tree parity (`decimal=8`) with legacy `n_reg=100`,
  `min_samples_treatment=10` for KL/ED/Chi — single-treatment (depths 1–3) and
  multi-treatment — plus a gate test showing `min_samples_treatment` prevents
  splits (and that the same config without it grows).
- The no-regularization parity tests from #946/#959 now pass `n_reg=0` /
  `min_samples_treatment=0` explicitly and still hold.
- Legacy `tests/test_uplift_trees.py` and `tests/test_causal_trees.py` — 47
  passed (no regression from the shared rebuild).

Part of #945. Closes #947.

### DIFF
--- a/causalml/inference/tree/_uplift/_builder.pxd
+++ b/causalml/inference/tree/_uplift/_builder.pxd
@@ -1,0 +1,10 @@
+# cython: cdivision=True
+# cython: boundscheck=False
+# cython: wraparound=False
+# cython: language_level=3
+
+from .._tree._tree cimport Node, Tree, TreeBuilder
+from .._tree._splitter cimport Splitter, SplitRecord
+from .._tree._typedefs cimport intp_t, int32_t, int64_t, float32_t, float64_t
+from .._tree._tree cimport FrontierRecord, StackRecord
+from .._tree._tree cimport ParentInfo, _init_parent_record
--- a/causalml/inference/tree/_uplift/_builder.pyx
+++ b/causalml/inference/tree/_uplift/_builder.pyx
@@ -1,0 +1,596 @@
+# distutils: language = c++
+# cython: cdivision=True
+# cython: boundscheck=False
+# cython: wraparound=False
+# cython: language_level=3
+
+"""Kernel tree builders for uplift trees with parent-shrinkage regularization.
+
+These mirror the stock ``DepthFirstTreeBuilder`` / ``BestFirstTreeBuilder`` in
+``_tree/_tree.pyx`` (same stopping rules, same ``__cinit__`` signatures) with two
+uplift-specific changes:
+
+1. **Parent-summary threading.** The regularization in
+   :class:`UpliftClassificationCriterion` shrinks each node toward its parent's
+   *regularized* ``P(Y=1|T=g)``. That summary is not derivable from the raw leaf
+   values stored in ``tree.value`` (which stay raw for prediction parity with the
+   legacy tree), so the builder keeps a ``node_id -> regularized summary`` map: it
+   sets the parent summary on the criterion before each ``node_split`` and stores
+   this node's regularized summary after ``node_value``, for its children to read.
+
+2. **No divergence-based early-leaf.** The stock builders turn a node into a leaf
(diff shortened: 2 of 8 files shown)


jeongyoonlee:

Issue #948, opened 2026-07-17
[Tree unification 3a/9] Rzepakowski normalization for KL/ED/Chi + CTS criterion


jeongyoonlee:
Add the Rzepakowski **normalization** factor (`normalization=True` is the default) to the KL/ED/Chi criteria, plus the **CTS** (Contextual Treatment Selection) criterion.

Depends on **Issue 1** (KL/ED/Chi scaffold). Grouped with Issue 3b (the two-class/variance criteria) under "remaining criteria".

## Scope

- Rzepakowski normalization `normI` / `arr_normI` (`uplift.pyx` ~1625–1734) applied to KL/ED/Chi so the gain is divided by the treatment-vs-control distribution divergence. Respect the `normalization` flag (default `True`).
- **CTS** criterion (simple multi-treatment gain: max over groups of the outcome mean).

## Verify

- Per-criterion parity vs legacy with `normalization=True` on fixed seeds (KL/ED/Chi).
- CTS parity vs `UpliftTreeClassifier(evaluationFunction="CTS")` on a fixed seed.

## Reusable templates

- Legacy: `uplift.pyx` `normI`/`arr_normI` (~1625–1734), `evaluate_CTS`.

Part of #945.

PR #962, opened 2026-07-23
Add kernel uplift normalization (arr_normI) and CTS criterion (#948)

jeongyoonlee:
**Stacked on #961 (issue #947) — review/merge #961 first.** Base is `feature/947-uplift-split-regularization`, so the diff below is only the #948 changes; GitHub retargets to `master` once #961 lands.

Part of the tree-unification epic #945 (child #948). Ports Rzepakowski split-gain **normalization** and the **CTS** criterion onto the kernel-backed uplift trees, keeping exact parity with the legacy `UpliftTreeClassifier`.

## What

- **`normalization`** (Rzepakowski et al. 2012, `arr_normI`, α=0.9): when enabled, the split gain is divided by `_norm_factor`, re-ranking splits by the treatment-vs-control distribution divergence. Refactors the per-node quantity behind a virtual `_node_metric` and adds a `_normalizable` hook so CTS can opt out (legacy never normalizes CTS).
- **`CTSCriterion`** (Contextual Treatment Selection, `evaluate_CTS`): node metric is the max over all groups of P(Y=1|T=g); the gain favors splits that raise the best achievable outcome. Wired in as the `"CTS"` criterion.

## Correctness note (the subtle bit)

`arr_normI` is **asymmetric** — it reads only *one* child's counts. Legacy's "left" branch is `col_vals >= value`, which is the sklearn splitter's **right** child, so `_norm_factor` reads `state.right`. Reading the left child passes every balanced-assignment parity test (where normalization never changes the chosen split) but selects the wrong split once treatment is correlated with a feature — which the new tests exercise.

## Tests

- Exact 8-decimal whole-tree parity with legacy for KL/ED/Chi with `normalization=True`, plus CTS single/multi-treatment, all at legacy defaults (`n_reg=100`, `min_samples_treatment=10`).
- A normalization-sensitive dataset + a parity test that *also* asserts normalization actually changes the fitted tree (guards a silent no-op or a wrong `_norm_factor`).
- Verified out-of-band over a 720-case sweep (40 seeds × KL/ED/Chi × depths 1–3 × norm on/off): 0 mismatches, 29 genuine norm-flips.
- 48 kernel + 47 legacy/causal tests pass locally.

Leaf predictions remain the raw `P(Y=1|T=g)` — normalization affects only the growth gain, matching legacy. Honesty, pruning, forest, and the public switchover follow in later epic children.

### DIFF
--- a/causalml/inference/tree/_uplift/_criterion.pxd
+++ b/causalml/inference/tree/_uplift/_criterion.pxd
@@ -18,6 +18,11 @@ cdef class UpliftClassificationCriterion(CausalRegressionCriterion):
     cdef public float64_t n_reg
     cdef public intp_t min_samples_treatment
 
+    # Rzepakowski et al. (2012) normalization: when set, the split gain is divided
+    # by the treatment-vs-control distribution divergence ``_norm_factor`` (issue
+    # #948). Not applied by criteria whose ``_normalizable`` returns 0 (e.g. CTS).
+    cdef public bint normalization
+
     # Regularized positive-outcome probabilities of the *parent* node, threaded
     # down by the builder before each ``node_split`` (control at index 0). Only
     # valid when ``has_parent`` is non-zero (the root has no parent).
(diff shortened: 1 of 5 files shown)


jeongyoonlee:

Issue #950, opened 2026-07-17
[Tree unification 4/9] Honesty (held-out leaf re-estimation) on the kernel Tree


jeongyoonlee:
Reimplement the **honesty** option (held-out leaf re-estimation) on the kernel `Tree`.

Depends on **Issue 1** (scaffold). Issue 3b's IDDP path depends on this landing.

## Scope

- Split the training data into a build split and an estimation split (`estimation_sample_size`, default 0.5).
- After the tree is grown on the build split, do a **post-build pass** re-estimating each leaf's per-group P(Y=1) on the estimation split — replacing the legacy `honestApproach` / `fillTree` — using `Tree.apply` to route estimation rows to leaves and rewrite each node's `value`.
- Respect `honesty=False` (default) as a no-op.

## Verify

- Honesty parity vs `UpliftTreeClassifier(honesty=True, ...)` on a fixed seed.
- IDDP path (which forces `honesty=True`) works end-to-end.

## Reusable templates

- Legacy: `uplift.pyx` `honestApproach` / `fillTree` (~615–883).
- Leaf routing: kernel `Tree.apply`.

Part of #945.

Issue #949, opened 2026-07-17
[Tree unification 3b/9] Two-class / variance criteria: DDP, IDDP, IT, CIT


jeongyoonlee:
Add the two-class / variance-based uplift criteria that build on the per-group variance and counts already available in the causal `NodeState`: **DDP**, **IDDP**, **IT**, **CIT**.

Depends on **Issue 1** (scaffold). Grouped with Issue 3a under "remaining criteria".

## Scope

- **DDP** — delta-delta-P: `effect(t)` = P(Y=1|T=t) − P(Y=1|control).
- **IDDP** — ΔΔP with `maxDiffSign` sign tracking; enforce the IDDP → `honesty=True` coupling (`uplift.pyx`). (Honesty itself lands in Issue 4; this issue wires the coupling + gain.)
- **IT** — squared t-statistic of the treatment effect, using `outcome_var(g)` and counts.
- **CIT** — likelihood-ratio test statistic.
- Enforce the **two-class-only** guards (`uplift.pyx` ~533–536) for DDP/IDDP/IT/CIT — raise a clear error on >1 treatment group.

## Verify

- Per-criterion parity vs legacy on fixed seeds (two-treatment fixture: `generate_classification_data_two_treatments`).
- Two-class guard raises on multi-treatment input.

## Reusable templates

- Legacy: `uplift.pyx` `evaluate_DDP`/`evaluate_IDDP`/`evaluate_IT`/`evaluate_CIT`, `maxDiffSign` tracking, two-class guards (~533–536).
- Variance accessors: causal `NodeState.outcome_var(g)`.

Part of #945.

PR #963, opened 2026-07-23
Add two-class uplift criteria DDP/IT/CIT to the kernel tree (#949)

jeongyoonlee:
**Stacked on #962 (issue #948) — review/merge #962 first.** Base is `feature/948-normalization`, so the diff below is only the #949 changes; GitHub retargets down the stack as the parents land.

Part of the tree-unification epic #945 (child #949). Ports the two-class / variance-based uplift split criteria onto the kernel-backed uplift trees, keeping exact parity with the legacy `UpliftTreeClassifier`.

## What

- **DDP** (delta-delta-P): gain = `|DDP(left) - DDP(right)|`, where `DDP(s) = Σ_t (P(Y=1|T=t) - P(Y=1|control))`.
- **IT** (interaction tree): squared t-statistic of the effect difference between the children, from each regularized rate's Bernoulli variance `p*(1-p)` and the raw group counts.
- **CIT** (causal inference tree): likelihood-ratio test statistic.

All three contrast a single treatment against control and are **never normalized**; a **two-class guard** raises on multi-treatment input (matching legacy `uplift.pyx` ~533-536).

## How

These gains aren't the `p·M_left + (1-p)·M_right - M_node` divergence form, so the gain is now computed by a virtual `_split_gain(s_node, s_left, s_right, p)` (the divergence/CTS criteria keep the default). IT/CIT need both regularized child summaries at once, so a second scratch buffer `_buf_right` was added. Leaf predictions remain the raw `P(Y=1|T=g)`.

## Tests

- Exact 8-decimal whole-tree parity vs legacy for DDP/IT/CIT at depths 1-3 (legacy defaults `n_reg=100`, `min_samples_treatment=10`), single treatment.
- A multi-treatment guard test (DDP/IT/CIT raise).
- Verified out-of-band over a 270-case sweep (30 seeds × 3 criteria × depths 1-3): 0 mismatches.
- 60 kernel + 47 legacy/causal tests pass locally.

## Deferred: IDDP → #950 (honesty)

Legacy **forces `honesty=True` for IDDP** (`uplift.pyx:468`), and the kernel has no honesty pass yet, so kernel-vs-legacy IDDP parity can't be verified until the honest leaf re-estimation lands. IDDP (its gain + second `arr_normI` normalization variant + the honesty coupling) will land in #950 where it's testable, rather than shipping unverifiable here.

### DIFF
--- a/causalml/inference/tree/_uplift/_criterion.pxd
+++ b/causalml/inference/tree/_uplift/_criterion.pxd
@@ -30,8 +30,11 @@ cdef class UpliftClassificationCriterion(CausalRegressionCriterion):
     cdef vector[float64_t] parent_summary_p
 
     # Scratch buffers (size n_outputs) for the regularized node / child summaries.
+    # ``_buf_left`` and ``_buf_right`` coexist so criteria that need both children
+    # at once (IT / CIT) can read them together.
     cdef vector[float64_t] _buf_node
-    cdef vector[float64_t] _buf_child
+    cdef vector[float64_t] _buf_left
+    cdef vector[float64_t] _buf_right
 
     # Divergence contribution of a single treatment-vs-control pair. Overridden
     # by each concrete criterion (KL / ED / Chi).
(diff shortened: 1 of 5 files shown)


jeongyoonlee:

Issue #955, opened 2026-07-17
[Tree unification 9/9] Switchover + delete legacy uplift.pyx + deprecations


jeongyoonlee:
Flip the public API to the kernel-backed implementation, delete the legacy `uplift.pyx` duplicate, and add deprecations. The final issue of the epic.

Depends on **all** prior issues (1–8) — full criterion coverage, regularization, normalization, honesty, pruning, forest, plot, serialization all at parity.

## Scope

- Flip public `UpliftTreeClassifier` / `UpliftRandomForestClassifier` to thin subclasses of the kernel-backed impls (names / args / defaults preserved).
- Delete the ~2.3 KLoC duplicated node / builder / splitter / bagging / scalar-criteria from `uplift.pyx`.
- Optionally rename `_uplift/` → `uplift/` (package) now that `uplift.pyx` is gone.
- Deprecation-warn on the public `DecisionTree` class (open Q2) and on any inevitable arg/behavior drift.

## Verify

- Full `tests/test_uplift_trees.py` green against the new impl.
- `tests/test_serialization_extended.py` green.
- `check_estimator` for both uplift classes.
- `uv run black .`.

## Reusable templates

- The kernel-backed impls from Issues 1–8; the causal classes as the structural precedent.

Part of #945.

Issue #952, opened 2026-07-17
[Tree unification 6/9] UpliftRandomForestClassifier on sklearn ForestClassifier


jeongyoonlee:
Reimplement `UpliftRandomForestClassifier` on top of sklearn's `ForestClassifier` scaffolding, mirroring `CausalRandomForestRegressor`, replacing the legacy joblib bagging forest.

Depends on the single-tree work (Issues 1–5) being complete enough that a single kernel-backed uplift tree matches legacy across criteria/regularization/honesty/pruning.

## Scope

- Base estimator = kernel-backed uplift tree; sklearn `ForestClassifier` scaffolding + a vendored `_parallel_build_trees` adapted for the `(X, treatment, y)` signature (mirror `causalforest.py`).
- Preserve the existing public behavior:
  - `_align_tree_predict` handling for bootstrap trees that are missing some treatment groups (the #569 sparse-group case),
  - `joblib_prefer`,
  - delta-column `predict` (`full_output`),
  - `feature_importances_` averaging.

## Verify

- The existing 12-case parametrized forest test (backends × prefer × early_stopping).
- The #569 sparse-group predict-shape regression.
- Parity vs legacy `UpliftRandomForestClassifier` on a fixed seed.

## Reusable templates

- Forest: `causalml/inference/tree/causal/causalforest.py` (`ForestRegressor` subclass, vendored `_parallel_build_trees`, `_fit`).
- Legacy: `uplift.pyx` forest (~2503–2742), `_align_tree_predict` (~64–82).

Part of #945.

Issue #569, opened 2022-11-08
shape mismatch when calling UpliftRandomForestClassifier.predict()


adsimone-aws:
**Describe the bug**
I am trying to fit/predict using a UpliftRandomForestClassifier.
According to the class documentation, the data passed to predict should have shape=(n_data_points, n_features)
Even when doing this, however, the call to predict fails with:

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-13-96a775b45bda> in <module>
     37                  y=test_outcome)
     38 
---> 39 uplift_model.predict(test_features)

~/.pyenv/versions/skynet/lib/python3.6/site-packages/sklearn/utils/_testing.py in wrapper(*args, **kwargs)
    306             with warnings.catch_warnings():
    307                 warnings.simplefilter("ignore", self.category)
--> 308                 return fn(*args, **kwargs)
    309 
    310         return wrapper

causalml/inference/tree/uplift.pyx in causalml.inference.tree.uplift.UpliftRandomForestClassifier.predict()

ValueError: operands could not be broadcast together with shapes (10,4) (10,3) 
```

**To Reproduce**
This is a very minimal setup, using only 10 data points. Predict is called on the train data. I see the same behavior on a much more complex setup. 
Please understand I had to hash the names of the treatment groups before publishing here.

```python
import numpy as np
from causalml.inference.tree import UpliftRandomForestClassifier

test_features = np.array([[235.9       , 229.9       , 229.9       , 223.9       ,
        239.21578947, 223.4       ],
       [235.9       , 229.9       , 229.9       , 223.9       ,
        238.40877193, 223.4       ],
       [235.9       , 229.9       , 239.9       , 223.9       ,
        238.97017544, 223.2       ],
       [235.9       , 229.9       , 239.9       , 223.9       ,
        239.77719298, 223.3       ],
       [235.9       , 229.9       , 239.9       , 221.9       ,
        239.68947368, 227.7       ],
       [239.9       , 229.9       , 239.9       , 221.9       ,
        240.28596491, 229.4       ],
       [239.9       , 229.9       , 249.9       , 221.9       ,
        242.46140351, 232.        ],
       [235.9       , 229.9       , 249.9       , 221.9       ,
        243.46140351, 234.3       ],
       [235.9       , 235.9       , 249.9       , 221.9       ,
        244.65438596, 235.4       ],
       [235.9       , 235.9       , 249.9       , 221.9       ,
[...]

PR #967, opened 2026-07-24
Add kernel-backed uplift random forest (#952)

jeongyoonlee:
## What

Reimplements `UpliftRandomForestClassifier` on scikit-learn's `ForestRegressor`
scaffolding as the experimental `_KernelUpliftRandomForestClassifier`, bagging
the kernel `_KernelUpliftTreeClassifier` (mirrors `CausalRandomForestRegressor`).
Part of the tree-unification epic #945 (issue #952, "6/9").

Not wired into the public API — it stays experimental alongside the kernel tree
until the epic switchover (#955).

## How

- **Bootstrap.** Legacy-style row resampling (so a bootstrap can miss whole
  treatment groups) via a vendored `_parallel_build_tree`, seeded from the same
  parent-RNG flow as the legacy forest.
- **`_align_tree_predict`.** Scatters each tree's per-group probabilities into
  the forest-wide class ordering (zero-filling absent groups) before averaging —
  the #569 sparse-group case. Degenerate control-only bootstraps carry no
  treatment signal and are dropped (only reachable under extreme imbalance; the
  legacy forest kept them as no-ops).
- **`predict`.** Returns per-treatment uplift deltas; `full_output=True` returns
  the per-group / recommended-treatment / delta frame. `joblib_prefer` preserved;
  `feature_importances_` inherited from `ForestRegressor` (per-tree averaging).
- No validation-set early stopping — the kernel tree does not implement it.

## Verify

New tests in `tests/test_uplift_trees_kernel.py`:

- **Whole-forest parity** vs the legacy `UpliftRandomForestClassifier`, bit-exact
  (`decimal=8`) on binary-feature data — KL/ED/Chi × normalization, plus a honest
  case. (Exact parity needs binary features + `legacy_max_depth = kernel + 1`;
  same conventions as the single-tree parity tests.)
- **#569** sparse-group predict-shape regression (1 sample per minority group):
  some tree misses a group, prediction is full-width and NaN-free, serial ==
  parallel.
- **Backend × `joblib_prefer`** determinism + AUUC > random (mirrors the legacy
  12-case test minus the early-stopping axis).
- `full_output` columns and `feature_importances_`.

Full kernel suite: 104 passed.

Part of #945.

### DIFF
--- a/causalml/inference/tree/_uplift/upliftforest.py
+++ b/causalml/inference/tree/_uplift/upliftforest.py
@@ -1,0 +1,232 @@
+"""Experimental kernel-backed uplift random forest classifier.
+
+Not part of the public API -- this bags the experimental
+``_KernelUpliftTreeClassifier`` on scikit-learn's ``ForestRegressor``
+scaffolding (mirroring ``CausalRandomForestRegressor``) to prove parity with
+the legacy joblib-bagging ``UpliftRandomForestClassifier`` before the public
+class is switched over.
+
+Like the legacy forest it bootstraps by resampling rows (so a tree can miss a
+treatment group entirely -- the #569 sparse-group case handled by
+``_align_tree_predict``), keeps the ``joblib_prefer`` knob, and predicts the
+per-treatment delta columns (``full_output`` returns the full frame). Early
+stopping on a validation set -- a legacy tree feature -- is not supported here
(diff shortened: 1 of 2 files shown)


ras44 (approved):
LGTM

Issue #805, opened 2024-11-27
plot_causal_tree can't show the significance level p


NancyLTY:
Now CausalTreeRegressor and trees in CausalRandomForestRegressor class can be plotted with the information about the treatment and control sample sizes that were used to build a tree. See from https://github.com/alexander-pv/causalml/blob/d1b841f64814bb01ce513c09f0ce98703a19572a/causalml/inference/tree/plot.py#L324, but this tree does not show the significance level p.

PR #976, opened 2026-07-27
#805: Add node p-value display to plot_causal_tree

genrichez:
## Proposed changes

Add the ability to display treatment effect p-values at each node when plotting a causal tree via `plot_causal_tree`. This addresses #805.

When `node_pvalues=True` is passed to `CausalTreeRegressor` during fit, a Welch's t-test is computed at each node comparing treatment vs control outcomes. The resulting p-values can then be rendered in the plot by passing `pvalue=True` to `plot_causal_tree`.

Usage:
```python
model = CausalTreeRegressor(control_name=0, groups_cnt=True, node_pvalues=True)
model.fit(X=X, treatment=treatment, y=y)
plot_causal_tree(model, pvalue=True)
```

Each node will display `p_value(group) = X.XXXX` for each treatment group. Nodes with fewer than 2 samples in either group show `N/A`.

## Types of changes

## Checklist

## Further comments

Implementation follows the same pattern as the existing `groups_cnt` feature:
- `node_pvalues=True` in the constructor enables computation during `fit()`
- `_compute_node_pvalues()` traverses the tree via `decision_path` (same approach as `_count_groups_distribution`)
- Uses Welch's t-test (`scipy.stats.ttest_ind`, `equal_var=False`) which is standard for unequal group sizes
- Multi-treatment support: computes a separate p-value for each treatment group vs control
- Gracefully handles nodes with insufficient samples (returns None)

This PR was co-authored with AI assistance.

### DIFF
--- a/causalml/inference/tree/causal/causaltree.py
+++ b/causalml/inference/tree/causal/causaltree.py
@@ -117,7 +117,10 @@ class CausalTreeRegressor(SerializableLearner, RegressorMixin, BaseCausalDecisio
                 See :term:`Glossary <random_state>` for details.
             groups_cnt: (bool), count treatment and control groups for each node/leaf
             groups_cnt_mode: (str, 'nodes', 'leaves'), mode for samples counting
-            node_pvalues: (bool), compute treatment effect p-values for each node
+            node_pvalues: (bool), compute treatment effect p-values for each node.
+                Note: These are naive in-sample t-tests and do not account for the tree
+                structure or multiple testing. They should be used for descriptive
+                purposes only and are not valid for post-selection inference.
         """
 
         self.criterion = criterion
@@ -464,6 +467,9 @@ class CausalTreeRegressor(SerializableLearner, RegressorMixin, BaseCausalDecisio
     ) -> dict:
         """
         Compute treatment effect p-values for each tree node using Welch's t-test.
+
+        Note: These p-values are descriptive and do not account for the search
+        process used to find the splits (no honesty or multiple-testing correction).
 
         Args:
             X: (np.ndarray), feature matrix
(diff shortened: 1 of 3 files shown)


jeongyoonlee (commented):
The node p-values are naive in-sample within-node t-tests — the partition is learned from the same data and splits maximize effect separation, so they're anti-conservative (no honesty/selective-inference or multiple-testing correction) and are raw diff-in-means (valid only under in-node unconfoundedness). Fine as the diagnostic #805 asks for, but please add a docstring note on `node_pvalues` / `plot_causal_tree(pvalue=...)` that these are descriptive, not valid post-selection inference.

Minor: a zero-variance node makes `ttest_ind` return `nan`, rendered as `p_value(g) = nan` — worth mapping to `N/A` like the insufficient-samples branch.

genrichez:
> The node p-values are naive in-sample within-node t-tests — the partition is learned from the same data and splits maximize effect separation, so they're anti-conservative (no honesty/selective-inference or multiple-testing correction) and are raw diff-in-means (valid only under in-node unconfoundedness). Fine as the diagnostic #805 asks for, but please add a docstring note on `node_pvalues` / `plot_causal_tree(pvalue=...)` that these are descriptive, not valid post-selection inference.
> 
> Minor: a zero-variance node makes `ttest_ind` return `nan`, rendered as `p_value(g) = nan` — worth mapping to `N/A` like the insufficient-samples branch.

Thanks for the thorough review! I've addressed these.
I've added the cautionary docstring notes for node_pvalues in both CausalTreeRegressor and plot_causal_tree, clarifying that these are descriptive p-values and not for post-selection inference ,also implemented the NaN to N/A mapping for p-values in the plot, and added a regression test to ensure this behavior is consistent.

jeongyoonlee (approved):

PR #977, opened 2026-07-29
CI: run optional backend tests on pull requests

### LINKED ISSUE #960: CI: optional-backend tests (TF/torch/JAX) are not run on PRs
### Problem

The optional-backend test suites (TensorFlow, PyTorch, JAX) are **not exercised by the PR CI**, so backend code can regress with a green pipeline.

- `.github/workflows/python-test.yaml` runs on PRs as `pytest -vs tests/ --cov causalml/` — with **no** `--runtf` / `--runtorch` / `--runjax`, so every backend test is skipped (see `tests/conftest.py`).
- The only workflow that passes those flags, `.github/workflows/test-build-from-source.yml`, triggers **only on `push` when `envs/*.yml` or that workflow file changes** — not on normal pull requests.

Net effect: the DragonNet JAX backend (#918), the CEVAE JAX backend (#957), the CEVAE torch backend, and the DragonNet TF backend all ship without their tests running in PR CI. The recent JAX CEVAE PR (#957) merged with its entire `tests/test_jax_cevae.py` suite skipped in CI (verified passing locally instead).

Additionally, the JAX↔pyro parity test (`test_cevae_jax_matches_pyro`) is marked with **both** `jax` and `torch`, so it needs `--runjax` **and** `--runtorch` in the same run. The `test-build-from-source.yml` matrix passes only one flag per row, so that test never runs anywhere in CI.

### Proposed fix

Add PR-triggered coverage for the optional backends, e.g.:

- Extend `python-test.yaml` (or add a dedicated `pull_request` job) that installs `.[test,tf,torch,jax]` and runs the backend flags. Per-backend jobs avoid in-process torch/JAX OMP conflicts; the CEVAE parity test already isolates each backend in its own subprocess, but it needs a job where **both** torch and jax are installed and **both** `--runjax --runtorch` are passed.
- Keep `test-build-from-source.yml` as the from-source/env smoke test.

### Impact

Low risk, CI-only. Prevents silent regressions in the TF/torch/JAX backends.

mohsinm-dev:
## Summary
- add isolated TensorFlow, PyTorch, JAX, and JAX-PyTorch parity jobs for Python 3.11 and 3.12
- install CPU-only PyTorch in CI and keep the existing from-source workflow focused on environment smoke testing
- repair and reduce the stale PyTorch CEVAE test so the newly enforced lane starts green

## Tests
- `pre-commit` hooks: YAML, EOF, trailing whitespace, Black
- `actionlint .github/workflows/optional-backend-tests.yml`
- TensorFlow: 1 passed
- PyTorch: 1 passed
- JAX: 6 passed
- JAX-PyTorch parity: 1 passed
- verified marker collection selects 1 TF, 1 PyTorch, 6 JAX-only, and 1 parity test

## Dependencies
- no package dependency changes
- CI uses the official PyTorch CPU wheel index to avoid CUDA packages on CPU runners

Closes #960

### DIFF
--- a/.github/workflows/optional-backend-tests.yml
+++ b/.github/workflows/optional-backend-tests.yml
@@ -1,0 +1,69 @@
+name: Optional backend tests
+
+on:
+  pull_request:
+  push:
+    branches:
+      - master
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: optional-backends-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  test:
+    name: ${{ matrix.backend }} (Python ${{ matrix.python-version }})
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.11", "3.12"]
+        backend: [tensorflow, pytorch, jax, jax-pytorch-parity]
+        include:
+          - backend: tensorflow
+            extras: tf
+            test-path: tests/
+            marker: tf
+            pytest-flags: --runtf
+          - backend: pytorch
+            extras: torch
+            test-path: tests/
+            marker: torch and not jax
+            pytest-flags: --runtorch
+          - backend: jax
+            extras: jax
+            test-path: tests/
+            marker: jax and not torch
+            pytest-flags: --runjax
+          - backend: jax-pytorch-parity
+            extras: torch,jax
+            test-path: tests/test_jax_cevae.py
+            marker: jax and torch
+            pytest-flags: --runjax --runtorch
+
+    steps:
+      - uses: actions/checkout@v6
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v6
+        with:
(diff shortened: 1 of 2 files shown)


jeongyoonlee (changes requested):
- Parity lane pins `test-path: tests/test_jax_cevae.py`, so a future test marked both `jax` and `torch` in any other file gets deselected by both single-backend lanes and runs nowhere — set it to `tests/`, since `-m "jax and torch"` already scopes it.
- Use `RANDOM_SEED` from `tests/const.py` in `tests/test_cevae.py` (`from .const import RANDOM_SEED`) instead of the literal `0` for both seed calls — it's 42, so please confirm the pytorch lane stays green.

mohsinm-dev:
Addressed both points:

- The JAX–PyTorch parity lane now scans `tests/` and relies on the `jax and torch` marker for selection.
- `tests/test_cevae.py` now uses the shared `RANDOM_SEED` for both NumPy and PyTorch.

The complete optional-backend matrix passes on Python 3.11 and 3.12, including both PyTorch and parity lanes.

jeongyoonlee (approved):
LGTM. Thanks!

PR #978, opened 2026-07-31
Skip one-hot-encoding in load_data when no feature is categorical

arpitjain099:
## Proposed changes

`load_data` raises when every requested feature is numeric:

```python
>>> load_data(pd.DataFrame({"a": [1.0, 2.0, 3.0], "b": [4, 5, 6]}), ["a", "b"])
AssertionError: no column was transformed, please check your dataframe input
```

`cat_cols` comes out empty, but `ohe.fit_transform(df[cat_cols])` is still called, and `OneHotEncoder` asserts that it transformed at least one column. An all-numeric feature matrix is a perfectly ordinary input, so this guards the one-hot branch and returns the numeric columns as they are when there is nothing categorical to encode.

The existing `test_load_data` uses `generate_categorical_data`, which always has categorical columns, so this path was not covered.

## Types of changes

## Checklist

## Further comments

`test_load_data_all_numeric_features` fails on master with the assertion above and passes with the change. `pytest tests/test_features.py` is 4 passed, and `black --check` is clean on both touched files.

### DIFF
--- a/causalml/features.py
+++ b/causalml/features.py
@@ -258,10 +258,16 @@ def load_data(data, features, transformations={}):
     cat_cols = [col for col in features if not pd.api.types.is_numeric_dtype(df[col])]
     num_cols = [col for col in features if col not in cat_cols]
 
-    logger.info("Applying one-hot-encoding to {}".format(cat_cols))
-    ohe = OneHotEncoder(min_obs=df.shape[0] * 0.001)
-    X_cat = ohe.fit_transform(df[cat_cols]).todense()
+    if cat_cols:
+        logger.info("Applying one-hot-encoding to {}".format(cat_cols))
+        ohe = OneHotEncoder(min_obs=df.shape[0] * 0.001)
+        X_cat = ohe.fit_transform(df[cat_cols]).todense()
 
-    X = np.hstack([df[num_cols].values, X_cat])
+        X = np.hstack([df[num_cols].values, X_cat])
+    else:
+        # OneHotEncoder asserts that at least one column was transformed, so
+        # skip it entirely when every feature is already numeric.
+        logger.info("No categorical features to one-hot-encode")
+        X = df[num_cols].values
 
     return X
--- a/causalml/features.py
+++ b/causalml/features.py
@@ -243,7 +243,7 @@ def load_data(data, features, transformations={}):
         transformation (dict of (str, func)): transformations to be applied to features
 
     Returns:
-        X (numpy.matrix): a feature matrix
+        X (numpy.ndarray): a feature matrix
     """
 
     df = data[features].copy()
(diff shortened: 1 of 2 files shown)


jeongyoonlee (changes requested):
- The new `else` branch returns `ndarray` while the existing path returns `numpy.matrix`, and `ndarray` is the right one to converge on — `load_data` is the only function in the package that produces a matrix (the sole `.todense()` is `features.py:263`, and every other documented return is an array), so please switch that line to `.toarray()`, update the `Returns:` at `features.py:246` to `numpy.ndarray`, and assert the type in the new test, since `shape` and `assert_array_equal` pass for either.
- `if cat_cols:` guards the wrong condition — `OneHotEncoder` asserts on "no column produced a non-empty encoding", and a single-valued categorical column encodes to label `0`, which `_transform_col` drops via `labels > 0` — so `load_data(pd.DataFrame({"const_cat": ["x"] * 7, "num1": [1, 2, 1, 2, 1, 1, 1]}), ["const_cat", "num1"])` still raises the same `AssertionError` on this branch; worth covering alongside the all-numeric case.

arpitjain099 (commented):
Both fixed in d56fbe7, thanks, you were right on both counts.

**Encoder output rather than `cat_cols`.** Your example does still raise on my first version, I reproduced it before changing anything. A single-valued column gets `offset = 0`, so every row maps to label 0, `_transform_col` drops those with `labels > 0` and returns `None`, and `transform` then asserts even though `cat_cols` was non-empty. The guard now sits on the encoder result: run the encoding when there are categorical columns, and treat "nothing came back" as no categorical block to append. I catch the `AssertionError` rather than trying to predict the outcome from `label_maxes`, since that would duplicate the encoder's own rule and could drift from it. Say the word if you would rather have the explicit check.

**ndarray.** `.todense()` is now `.toarray()`, the `Returns:` says `numpy.ndarray`, and each test asserts `isinstance(features, np.ndarray) and not isinstance(features, np.matrix)`, since as you note `shape` and `assert_array_equal` are happy either way. I checked the one in-tree caller, `match.py:527`, which passes the result straight to `ElasticNetPropensityModel.fit_predict`, so it takes the ndarray without any change.

Tests are now three: all-numeric, your constant-categorical case, and a categorical case asserting the return type. The last two fail on the previous commit, all six in the file pass now, and `black --check` is clean.

jeongyoonlee (approved):
One non-blocking note.

`except AssertionError` uses an assertion as control flow, so the guard is compiled out under `python -O` / `PYTHONOPTIMIZE=1` — `transform` returns `None` rather than asserting, and the constant-categorical case this PR adds then raises `AttributeError: 'NoneType' object has no attribute 'toarray'`. Moving `.toarray()` out of the `try` closes that without re-deriving the encoder's rule you didn't want to duplicate:

```python
        try:
            X_cat = ohe.fit_transform(df[cat_cols])
        except AssertionError:
            logger.info("No categorical column produced an encoding")

        if X_cat is not None:
            X_cat = X_cat.toarray()
```

Happy to take it in this PR or a follow-up.

arpitjain099:
Good catch, and it reproduces exactly as you described. On master, `python -O` gives `AttributeError: 'NoneType' object has no attribute 'toarray'` for the constant-categorical input, since the stripped assertion means `transform` returns `None` and the `except` never fires.

Took it as a follow-up in #979 with your suggested shape, `fit_transform` inside the `try` and `.toarray()` outside behind a `None` check, so nothing re-derives the encoder's rule.

The test there monkeypatches `fit_transform` to return `None`, which reproduces optimized mode deterministically without a second interpreter. It fails on master with that `AttributeError`. I also ran the real `python -O` both ways to be sure, raises before, returns a `(7, 1)` array after.

Thanks for the review on this one, both notes were right.

PR #979, opened 2026-07-31
Handle a None encoding in load_data when assertions are disabled

arpitjain099:
## Proposed changes

Follow-up to #978, taking @jeongyoonlee's note from that review.

`OneHotEncoder.transform` signals "no column produced a non-empty encoding" with an `assert`, and `load_data` catches it. Under `python -O` (or `PYTHONOPTIMIZE=1`) that assertion is compiled out, so `transform` returns `None` instead of raising, the `except` never runs, and `.toarray()` lands on `None`:

```
$ python -O -c "from causalml.features import load_data; ..."
AttributeError: 'NoneType' object has no attribute 'toarray'
```

That is the same constant-categorical input the assertion path already handles, so it works normally and fails only when assertions are off.

The fix keeps `fit_transform` in the `try` and moves `.toarray()` out behind a `None` check, so both signals converge on the same branch. No re-deriving of the encoder's own rule.

## Types of changes

## Checklist

## Further comments

The test monkeypatches `OneHotEncoder.fit_transform` to return `None`, which reproduces the optimized-mode behavior deterministically without spawning a second interpreter. It fails on current master with the `AttributeError` above and passes with this change.

I also ran the real thing both ways. On master, `python -O` raises; with this patch it returns a `(7, 1)` array. `pytest tests/test_features.py` is 7 passed and `black --check` is clean.

### DIFF
--- a/causalml/features.py
+++ b/causalml/features.py
@@ -263,13 +263,19 @@ def load_data(data, features, transformations={}):
         logger.info("Applying one-hot-encoding to {}".format(cat_cols))
         ohe = OneHotEncoder(min_obs=df.shape[0] * 0.001)
         try:
-            X_cat = ohe.fit_transform(df[cat_cols]).toarray()
+            X_cat = ohe.fit_transform(df[cat_cols])
         except AssertionError:
             # OneHotEncoder asserts when no column produced a non-empty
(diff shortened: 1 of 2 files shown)


jeongyoonlee (approved):
LGTM

Issue #985, opened 2026-07-31
Flip every signature to `(X, y, treatment, ...)` and delete the deprecation shim


jeongyoonlee:
Part of #980 (v1.0 M1). Refs #854.

The breaking change itself: reorder every signature to `(X, y, treatment, …)` and delete the deprecation shim.

**This lands as one change.** Once a signature is reordered, existing positional calls break *silently* — `y` and `treatment` are both same-length arrays, so there is no `TypeError` to catch a swap. A partially flipped package is therefore worse than either end state, and no subset of this checklist may be released on its own.

**Unblocked and scheduled.** The window-closing issues are merged (#981 shim coverage, #982 tests, #983 docs, #984 migration guide), and the release schedule is fixed: the warning ships in 0.18.0 (Sep 2026), and this flip lands in **v1.0 (Jun 2027)** — three minor releases, roughly nine months.

## The rule

`X`, then `y`, then `treatment`; every other parameter keeps its relative position. Already implemented as `v1_order()` in `causalml/inference/_arg_order.py` and derived per method from that method's own signature — so the target for each method is machine-readable, not hand-maintained. `tests/test_fit_arg_order.py::test_v1_order` pins it.

Two shapes that are not a plain swap of the leading pair, both settled and already published in `docs/migration.rst`:

- A suffixed pair is reordered in place, so `UpliftTreeClassifier.fit` → `(X, y, treatment, X_val, y_val, treatment_val, sample_weight, check_input)`.
- `Sensitivity.get_*` takes `p` second: `(X, p, treatment, y)` → `(X, y, treatment, p)`.

## Scope

The authoritative list is **every method carrying `_arg_order_shimmed`** — 56 on a default install, plus 12 on the optional TF/torch/JAX backends. The window is one-shot, so each of those methods has already told users its v1.0 order, and each must arrive at exactly that order here. The checklist below is that set grouped by family:

- [ ] **Meta-learners** — `BaseSLearner`, `BaseTLearner`, `BaseXLearner`, `BaseRLearner`, `BaseDRLearner` and subclasses (`XGBTRegressor`, `XGBRRegressor`, `XGBRClassifier`, `XGBTClassifier`, `BaseSClassifier`, `BaseTClassifier`, `BaseXClassifier`, `BaseRClassifier`, `BaseDRClassifier`, `LRSRegressor`), plus `bootstrap` / `fit_bootstrap_ensemble`
[...]

Issue #982, opened 2026-07-31
Migrate `tests/` to keyword `fit`/`predict` calls (917 FutureWarnings)


jeongyoonlee:
Part of #980 (v1.0 M1). Refs #854.

After #975 the full test suite emits **917 `FutureWarning`s**, essentially all of them from CausalML's own tests calling `fit`/`predict` positionally. 11 files under `tests/` are affected.

Migrating them to keyword arguments — `learner.fit(X, y=y, treatment=treatment)` — does three things:

1. **Silences the noise**, so a genuinely new warning is visible in CI output instead of buried.
2. **Rehearses the migration.** These are the same call sites that break at the flip; converting them now proves the keyword form works everywhere before any signature moves.
3. **Makes the flip a near-no-op internally**, shrinking the blame surface if something regresses.

The change is mechanical and behaviour-preserving: keyword calls are order-independent, so each edit is verifiable by the test still passing.

## Acceptance

- `pytest tests/` emits zero arg-order `FutureWarning`s.
- No behavioural change: the suite passes with the same counts (366 passed / 10 skipped as of `88cde2f`).
- Deliberate exceptions are the tests in `tests/test_fit_arg_order.py` that *assert* the warning fires — those keep their positional calls.

Issue #980, opened 2026-07-31
Adopt the scikit-learn `fit(X, y, treatment, ...)` argument order (v1.0 M1)


jeongyoonlee:
## Summary

CausalML learners take `fit(X, treatment, y, ...)`, which puts `y` third. `sklearn.pipeline.Pipeline` calls the final estimator's `fit(X, y)` positionally, so today a CausalML learner can't be a Pipeline step without an adapter — the problem reported in #854.

The [v1.0 roadmap RFC](https://github.com/uber/causalml/discussions/938) lists this under **M1 — Foundation & stability** as **the one intentional breaking change**: adopt `fit(X, y, treatment, ...)`.

Reordering two required positional arguments can't be done in one step. `y` and `treatment` are both same-length arrays, so a straight swap would silently mis-train existing positional callers with no `TypeError` to catch it. So this is a **two-step deprecation**, and this epic tracks it end to end.

**Step one has landed** (#975): positional order is unchanged, but passing `treatment`/`y` positionally emits a `FutureWarning` pointing callers at keyword arguments — `fit(X, y=y, treatment=treatment)` — which are order-independent and therefore safe across the flip.

## The rule

One rule, applied uniformly: **`X`, then `y`, then `treatment`, and every other parameter keeps its relative position.** It is implemented in `causalml/inference/_arg_order.py` as `v1_order()`, derived per method from that method's own signature, so the target order is machine-checkable rather than hand-maintained:

| method | today | v1.0 |
|---|---|---|
| meta-learner `fit` | `(X, treatment, y, p, …)` | `(X, y, treatment, p, …)` |
| `predict` | `(X, treatment, y, p, …)` | `(X, y, treatment, p, …)` |
| `CausalTreeRegressor.fit` | `(X, treatment, y, sample_weight, …)` | `(X, y, treatment, sample_weight, …)` |
| `IVRegressor.fit` | `(X, treatment, y, w)` | `(X, y, treatment, w)` |
| `BaseDRIVLearner.fit` | `(X, assignment, treatment, y, …)` | `(X, y, treatment, assignment, …)` |

`predict` is included deliberately. Its current order doesn't block `Pipeline`, since `predict(X)` already works — but flipping only `fit` would leave v1.0 with the same two arguments in opposite positional orders on one class. Methods that don't take `treatment`/`y` at all, such as `BaseRLearner.predict`, are untouched.

## Why the sequencing matters

[...]

Issue #854, opened 2025-10-24
CausalML is not compatible with Scikit-Learn Pipeline


PelinV:
**Describe the solution you'd like**
Swap the arguments treatment and y in the fit function.
To change the order of arguments in the fit function from X, treatment, y to X, y, treatment for compatibility with sklearn

**Additional context**
When using causalml models in the sklearn pipeline, it is necessary to write adapters
Because pipeline expects the following order: X, y, other arguments, and in fit in causalml the following order: X, treatment, y

jbbqqf:
Hi! Triaging older issues — I think this one is a candidate for closure as out-of-scope.

Reordering `fit(X, treatment, y) → fit(X, y, treatment)` would break every meta-learner's public API (`BaseTLearner.fit`, `BaseRLearner.fit`, `BaseXLearner.fit`, etc. all use the `(X, treatment, y, ...)` signature; see `causalml/inference/meta/base.py:16` and corresponding lines in each subclass). Every example notebook, every downstream wrapper, every existing test would have to change. That's a major-version bump that the maintainers would need to schedule, not a drive-by fix.

The standard escape hatch for sklearn `Pipeline` compatibility is a small adapter that swaps the argument order (a few lines of code per learner). If the discussion is "should `causalml` ship that adapter," reframing as a fresh `meta: add SklearnPipelineCompat wrapper` issue with the proposed API would be cleaner than reopening the breaking-rename ask.

(Disclosure: I drafted this comment with help from Claude Code while triaging stale issues; the references above were verified manually against the current `master` branch.)

jeongyoonlee:
Thanks @PelinV — we're doing this. The `fit()` argument-order change is on the [v1.0 roadmap](https://github.com/uber/causalml/discussions/938) as the one intentional breaking change, moving the meta-learners to the scikit-learn convention `fit(X, y, treatment, ...)`.

Reordering two required positional arguments can't be done in one shot without risking a **silent** `y`/`treatment` swap for existing positional callers, so we're rolling it out as a two-step deprecation:

**Step 1 — #975 (open, non-breaking):** the positional order stays `(X, treatment, y)`, but passing `treatment`/`y` positionally now raises a `FutureWarning` steering you to keyword arguments:

```python
learner.fit(X, y=y, treatment=treatment)   # order-independent, silent, future-proof
```

Keyword calls are unaffected by the eventual flip, so migrating now is a one-time change.

[...]

PR #988, opened 2026-07-31
Stop the #854 shim warning on causalml's own internal fit calls

jeongyoonlee:
## Proposed changes

Follow-up to #975. The deprecation shim suppresses re-entrant warnings with a flag stored on the **instance**, so it covers `self.fit` → `self.predict` nesting but not the **forest → tree** hop: the flag is set on the forest while the warning fires on each tree it fits.

The result is a warning the caller cannot act on or silence:

```python
>>> CausalRandomForestRegressor(n_estimators=2).fit(X=X, y=y, treatment=treatment)
FutureWarning: Passing `treatment` and/or `y` to fit() by position is deprecated ...
FutureWarning: Passing `treatment` and/or `y` to fit() by position is deprecated ...
```

That call is all-keyword — written exactly the way the warning asks for — and it emits **one FutureWarning per tree**: 100 at the default `n_estimators`. `UpliftRandomForestClassifier` behaves the same way (default 10).

This makes the forests pass `y` and `treatment` to their trees by keyword. It is the same migration #982 applies to `tests/`, applied to the library's own call sites — the epic (#980) has no ticket covering those.

### Call sites changed

| File | |
|---|---|
| `causalml/inference/tree/causal/causalforest.py:136` | subsample / bootstrap branch |
| `causalml/inference/tree/causal/causalforest.py:145` | non-subsample branch |
| `causalml/inference/tree/_uplift/upliftforest.py:73` | bagged uplift tree fit |

### Verification

An empirical probe over eleven public entry points (both forests, both trees, the S/T/X/R/DR learners, `estimate_ate`, `fit_predict`), all called with keywords, attributes each warning to its originating library line via `stacklevel=2`: **2 offending sites before, 0 after**.

- `tests/test_fit_arg_order.py`: 54 passed.
- `tests/test_causal_trees.py tests/test_uplift_trees.py tests/test_uplift_trees_kernel.py tests/test_serialization_extended.py`: 135 passed.
- Both new regression tests fail with the library change reverted and pass with it, so they pin the fix rather than passing vacuously.
- `black --check` clean.

### Left alone, deliberately

The same-instance `super().fit(X, treatment, y, ...)` calls in `upliftforest.py:342`, `uplifttree.py:567` and `rlearner.py:913`. The guard already covers them and they warn nobody today. **They do need to change under #985** — once the signature order flips, `treatment` silently lands in `y`.

### Note on why this was not caught

The repo sets no `filterwarnings` anywhere, so `FutureWarning` is never an error and the affected tests pass on a normal run. A deprecation shim without a warnings-as-errors lane is untested by construction; worth pairing with the CI work in #977.

## Types of changes

## Checklist

### DIFF
--- a/causalml/inference/tree/_uplift/upliftforest.py
+++ b/causalml/inference/tree/_uplift/upliftforest.py
@@ -70,7 +70,7 @@ def _parallel_build_tree(tree, X, treatment, y, control_name, sample_weight):
     if control_name not in groups or len(groups) < 2:
         return None
     sw = None if sample_weight is None else sample_weight[idx]
-    tree.fit(X[idx], t_sub, y[idx], sample_weight=sw, check_input=True)
+    tree.fit(X[idx], y=y[idx], treatment=t_sub, sample_weight=sw, check_input=True)
     return tree
 
 
--- a/causalml/inference/tree/causal/causalforest.py
+++ b/causalml/inference/tree/causal/causalforest.py
@@ -135,8 +135,8 @@ def _parallel_build_trees(
 
         tree.fit(
             X,
-            treatment,
-            y,
+            y=y,
+            treatment=treatment,
             sample_weight=curr_sample_weight,
             check_input=True,
             prepare_data=False,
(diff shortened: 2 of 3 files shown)


paullo0106 (approved):
lgtm

PR #994, opened 2026-08-09
Add bootstrap confidence intervals to auuc_score() and qini_score()

ipezygj:
## What

`auuc_score()` and `qini_score()` return a bare point estimate. The two numbers people quote from an uplift model carry no indication of how much of them is sampling noise, so a Qini from a ranking built out of noise reads exactly like a Qini from a model that works:

```
>>> qini_score(df)
informative_model    0.1287
useless_model        0.0093
```

`useless_model` is a column of uniform random numbers. On the page, 0.0093 is a positive Qini.

This adds the same `return_ci` / `n_bootstrap` / `alpha` / `random_state` arguments that `rate_score()` already has:

```
>>> qini_score(df, return_ci=True, random_state=42)
                     qini      se  ci_lower  ci_upper  p_value
model
informative_model  0.1287  0.0120    0.1052    0.1523   0.0000
useless_model      0.0093  0.0095   -0.0093    0.0280   0.3271

>>> auuc_score(df, return_ci=True, random_state=42)
                     auuc      se  ci_lower  ci_upper
model
informative_model  0.6201  0.0034    0.6134    0.6268
useless_model      0.5068  0.0044    0.4981    0.5154
```

Default is `return_ci=False`, so nothing existing changes.

## Following the conventions already in the repo

- The keyword names and the returned frame match `rate_score(return_ci=True)` exactly — `score, se, ci_lower, ci_upper, p_value`, indexed by `model`. `return_ci` is already the house spelling in `BaseTLearner` (#886) and in sensitivity (#758) too.
- The resample is the same **half-sample bootstrap** (`m = n // 2`, without replacement) that `rate_score()` uses, for the same reason: these scores are functionals of a *ranking*, where the m-out-of-n resample stays valid and the naive n-out-of-n resample of a non-smooth functional need not.
- The bootstrap calls the scoring function itself on each resample, so the interval can never drift away from the point estimate it surrounds.

## One deliberate asymmetry: AUUC gets no p-value

`qini_score()` reports a p-value against H0 = 0 and `auuc_score()` does not, and that is not an oversight.

The Qini score is already the area *between* the model curve and the random curve, so zero means "no better than random" and is the right null. AUUC is not differenced against anything: a random ranking traces the diagonal, so its normalized AUUC is about 0.5. A two-sided test of H0: AUUC = 0 would reject for essentially every model ever scored and would tell a reader nothing about whether the model beats random.

Measured rather than asserted — twelve independently generated datasets, scoring a pure-noise ranking:

| score | mean over 12 datasets of a random ranking |
|---|---:|
| `qini_score` | **+0.0001** |
| `auuc_score` | **+0.5001** |

Both numbers are pinned by tests (`test_qini_null_really_is_zero`, `test_auuc_reports_no_p_value_because_its_null_is_not_zero`), so if the baseline ever moves, the test says so rather than the docstring quietly becoming wrong.

`return_ci=True` with `tmle=True` raises instead of running: each draw would refit the TMLE learner. The message points at `plot_tmlegain(ci=True)` / `plot_tmleqini(ci=True)`, which is where the TMLE path already has its own interval support.

## Testing

26 tests in `tests/test_visualize_score_ci.py`. The ones doing the real work:

- **negative control** — a noise ranking must come back with `p > 0.05` and a CI covering zero;
- **planted signal** — an informative ranking must come back separated from zero, so the control above cannot be passing merely for lack of power;
- **scaling** — quadrupling the sample must roughly halve the standard error, which catches a standard error computed on the wrong `n`;
- **the point estimate is untouched** — adding an interval must not move the number it is an interval around;
- reproducibility from `random_state`, argument validation, and the `tmle` refusal.

Rather than list assertions, I ran the suite against six deliberately broken versions of the implementation to check the tests can actually fail:

| mutation | result |
|---|---|
| no subsampling (`m = n`, every draw identical) | 4 failed |
| AUUC given a p-value against the wrong null | 2 failed |
| CI bounds swapped | 3 failed |
| argument validation disabled | 2 failed |
| `tmle` guard removed | 1 failed |
| `random_state` ignored | 2 failed |

Each mutation fails the tests it should. `black` is clean on both files.

Backward compatibility was checked rather than assumed: the new arguments are **keyword-only**, declared after `*args`, so nothing that currently reaches `*args`/`**kwarg` on the way to `get_tmlegain` / `get_tmleqini` changes binding. I also grepped every call site in the repo and the example notebooks — all of them pass `df` positionally and everything else by keyword.

**Limitation I should state plainly:** I could not build the full `causalml` package locally (Python 3.14 here, and the Cython extensions do not compile), so these tests were exercised against the real `visualize.py` scoring code loaded directly, with `lightgbm` and the TMLE learner stubbed — neither is on the code path being tested. They import only numpy, pandas, pytest and `causalml.metrics.visualize`, so they should be unremarkable in CI, but I have not watched them run there.

## A hypothesis I checked and dropped

Before writing this I suspected `rate_score()`'s half-sample standard error was missing a `sqrt(m/n)` factor and was therefore about 41% too wide. I measured it before saying anything: 300 independently generated datasets to get the true sampling spread of the score at `n`, compared against the half-sample bootstrap spread on a single dataset.

The ratios came out **0.87–1.07**, not 1.41. There is no missing factor — `rate_score()`'s standard error is on the right scale, and this PR copies its scheme unchanged. Recording the check here so nobody has to re-run it.

## On the issue-first guideline

CONTRIBUTING asks for an issue before work outside the existing issue list. I opened with the patch because it is small, additive and easier to judge as code than as a proposal — but if you would rather discuss the shape first, say so and I will move it to an issue without complaint. Closing this is a fine outcome too; the measured note about AUUC's 0.5 baseline is worth having on record either way.

## Follow-up, if wanted

`rate_score()` could share the `_bootstrap_score_ci()` helper this PR adds instead of keeping its own inline copy of the loop. I left that alone to keep the diff to one file and one behaviour change — happy to do it here or separately.

### DIFF
--- a/causalml/metrics/rate.py
+++ b/causalml/metrics/rate.py
@@ -36,6 +36,92 @@ def _compute_rate_from_toc(toc, weighting):
     )
 
 
+def _validate_bootstrap_args(n_bootstrap, alpha):
+    """Reject bootstrap settings that would produce an interval meaning nothing.
+
+    ``n_bootstrap`` must be at least 2: the standard error is ``np.std(..., ddof=1)``
+    over the draws, which is NaN for a single draw, and a NaN standard error would
+    otherwise travel onward as NaN interval bounds beside a confident-looking p-value.
+    """
+    if n_bootstrap < 2:
+        raise ValueError(
+            "n_bootstrap must be at least 2 to estimate a standard error, got {}".format(
+                n_bootstrap
+            )
+        )
+    if not 0 < alpha < 1:
+        raise ValueError(
+            "alpha must lie strictly between 0 and 1, got {}".format(alpha)
+        )
+
+
+def _bootstrap_score_ci(
+    df, score_fn, point, score_name, n_bootstrap, alpha, random_state, p_value=True
+):
+    """Half-sample bootstrap interval for a curve-summary score.
+
+    Draws m = n // 2 units without replacement, exactly as ``rate_score()`` does,
+    and for the same reason: these scores are functionals of a *ranking*, and the
+    m-out-of-n bootstrap stays valid where the naive n-out-of-n resample of a
+    non-smooth functional need not.
+
+    ``score_fn`` is the scoring function itself, called on each resample, so the
+    bootstrap can never drift from the point estimate it is an interval around.
+
+    Args:
+        df (pandas.DataFrame): the data the point estimate was computed on
+        score_fn (callable): maps a data frame to a Series of scores per model
+        point (pandas.Series): the point estimates
+        score_name (str): name for the point-estimate column in the result
+        n_bootstrap (int): number of half-sample draws
+        alpha (float): significance level
+        random_state (int or None): seed for the resampler
+        p_value (bool, optional): whether H0 = 0 is a meaningful null for this
+            score. True for Qini, which is already measured against random.
+            False for AUUC, whose random baseline is about 0.5 rather than 0.
+
+    Returns:
+        (pandas.DataFrame): score, se, ci_lower, ci_upper and, when meaningful,
+            p_value, indexed by model
+    """
+    n = len(df)
+    m = n // 2
+    rng = np.random.default_rng(random_state)
+    boot_scores = {model: [] for model in point.index}
+
+    for _ in range(n_bootstrap):
+        idx = rng.choice(n, size=m, replace=False)
+        resampled = score_fn(df.iloc[idx].reset_index(drop=True))
+        for model in point.index:
+            boot_scores[model].append(resampled[model])
+
+    z_crit = stats.norm.ppf(1 - alpha / 2)
+    results = []
+    for model in point.index:
+        estimate = point[model]
+        se = np.std(np.array(boot_scores[model]), ddof=1)
+        row = {
+            "model": model,
+            score_name: estimate,
+            "se": se,
+            "ci_lower": estimate - z_crit * se,
+            "ci_upper": estimate + z_crit * se,
+        }
+        if p_value:
+            # A p-value is only defined when the standard error is a real, positive
+            # number. Where it is not — a degenerate resample, or every draw landing
+            # on the same value — the honest answer is NaN. Dividing by a zero or NaN
+            # standard error would report p = 0.0 next to NaN interval bounds, which
+            # reads as certainty produced by the absence of a measurement.
+            if np.isfinite(se) and se > 0:
+                z_stat = estimate / se
+                row["p_value"] = 2 * (1 - stats.norm.cdf(abs(z_stat)))
+            else:
+                row["p_value"] = np.nan
+        results.append(row)
+
+    return pd.DataFrame(results).set_index("model")
+
 def get_toc(
     df,
     outcome_col="y",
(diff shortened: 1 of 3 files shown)


jeongyoonlee (changes requested):
**Blocking**
- `_bootstrap_score_ci`: `z_stat = estimate / se if se > 0 else np.inf` reports `p_value = 0.0` next to NaN CI bounds whenever `se` is NaN or 0 — reachable with `n_bootstrap=1`, which passes `_validate_bootstrap_args` while `np.std(ddof=1)` on a single draw is NaN, and with small `n`. Require `n_bootstrap >= 2`, and return `np.nan` for the p-value unless `se` is finite and positive; `causalml/metrics/rate.py:304` carries the same line.

**Non-blocking**
- Both `Returns:` blocks say `(float)` for the `return_ci=False` path; both functions return a `pd.Series` indexed by model column.
- `tests/test_visualize_score_ci.py`: use `from .const import RANDOM_SEED` — the `try`/`except ImportError` fallback to a bare `42` decouples the test from the shared constant.
- Take the `_bootstrap_score_ci` unification you offered — `rate_score`'s inline copy has already drifted, since only the new path validates `n_bootstrap`/`alpha`.

ipezygj:
Thank you — the blocking one is a real defect and you described it exactly. Fixed in `9a96603`.

**Blocking.** `n_bootstrap=1` passed validation while `np.std(ddof=1)` over a single draw is NaN, so the NaN reached the interval bounds while the p-value came from `estimate / se if se > 0 else np.inf` and printed `0.0`. A certain-looking answer produced by the absence of a measurement is the worst failure mode this PR could have shipped, given what it is for.

- `_validate_bootstrap_args` now requires `n_bootstrap >= 2`, with the reason in the docstring.
- The p-value is `np.nan` unless the standard error is finite and positive.

Two regression tests lock both: a single draw raises, and a degenerate resample where every draw lands on the same value returns `se == 0` with a NaN p-value rather than `0.0`.

**`rate.py:304`.** You are right that it carried the same line, and I would rather not fix it twice. The inline copy is gone: `_validate_bootstrap_args` and `_bootstrap_score_ci` now live in `rate.py` — it owns the half-sample rationale and carries no heavy imports, so the dependency points `visualize -> rate` rather than the other way — and `rate_score()` calls the shared implementation. It also validates its arguments now, which the inline copy never did.

**Docstrings.** Both `return_ci=False` paths now say `pandas.Series` indexed by model column.

**Test constant.** `from .const import RANDOM_SEED`, fallback removed.

One note on verification: `shap` is not installed in my environment, so `import causalml.metrics` fails before reaching this code and I could not run the suite end to end locally. I verified the two fixed behaviours by loading `rate.py` directly and exercising `_validate_bootstrap_args` and `_bootstrap_score_ci` against a constant scorer and a varying one; all three files compile. CI here is the real check, so please treat my local run as partial.

jeongyoonlee (changes requested):
**Blocking**
- The `black` check fails on `causalml/metrics/rate.py:124`.

**Non-blocking**
- `causalml/metrics/visualize.py:7`: `from scipy import stats` is unused.
- `rate.py:63`: "exactly as `rate_score()` does" reads oddly in the helper that is now `rate_score()`'s own implementation.

ipezygj:
Fixed — black was missing a second blank line before `get_toc` in `causalml/metrics/rate.py`. Formatting check passes locally on all three changed files.

ipezygj:
Both non-blocking points fixed in f5cf006 — the unused scipy import is gone and the docstring no longer refers to rate_score() as if it were a separate implementation.

jeongyoonlee (approved):

PR #999, opened 2026-08-10
Clamp the treatment probability in _kl_divergence

arpitjain099:
`FilterSelect._kl_divergence` guards the control probability against 0 and 1 but not the treatment one:

```python
if qk < 0.1**6:
    qk = 0.1**6
elif qk > 1 - 0.1**6:
    qk = 1 - 0.1**6
S = pk * np.log(pk / qk) + (1 - pk) * np.log((1 - pk) / (1 - qk))
```

When a bin contains no converters in the treatment arm, `pk` is 0 and the first term evaluates `0 * log(0 / qk)`, which numpy returns as NaN rather than the 0 the limit gives. `_evaluate_KL` adds that into the node total, `_filter_D_one_feature` adds the node totals together, and the feature's score and rank both come out NaN.

It is easy to hit on sparse conversion data. With a feature whose lower bins have no converters in either arm:

```
  feature method  score p_value               misc  rank
0       x     KL    NaN    None  number_of_bins: 5   NaN
```

plus `RuntimeWarning: divide by zero encountered in log` from line 384.

The change applies the same epsilon clamp to `pk`. With it, the same frame scores normally:

```
  feature method    score p_value               misc  rank
0       x     KL  0.00092    None  number_of_bins: 5   1.0
```

`ED` and `Chi` are unaffected since neither takes a log.

Verification: `tests/test_feature_selection.py::test_filter_kl_bin_with_no_conversions` is new and fails on master (`1 failed`), passes with the change. I ran it with `--noconftest` because the fixtures in `tests/conftest.py` pull in the compiled tree extensions, which are not built in my checkout. `black --check` is clean on both files.

### DIFF
--- a/causalml/feature_selection/filters.py
+++ b/causalml/feature_selection/filters.py
@@ -377,10 +377,15 @@ class FilterSelect:
             pk (float): Probability of class 1 in treatment group
             qk (float): Probability of class 1 in control group
         """
-        if qk < 0.1**6:
-            qk = 0.1**6
-        elif qk > 1 - 0.1**6:
-            qk = 1 - 0.1**6
+        eps = 0.1**6
+        if qk < eps:
+            qk = eps
+        elif qk > 1 - eps:
+            qk = 1 - eps
+        if pk < eps:
+            pk = eps
+        elif pk > 1 - eps:
+            pk = 1 - eps
         S = pk * np.log(pk / qk) + (1 - pk) * np.log((1 - pk) / (1 - qk))
         return S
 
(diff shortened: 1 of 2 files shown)


jeongyoonlee (approved):
**Non-blocking**
- `causalml/inference/tree/_uplift/_criterion.pyx:83` already implements this same scalar with exact branches — `pk == 0` → `-log(1 - qk)`, `pk == 1` → `-log(qk)` — which is the same amount of code as the clamp and exactly right; the clamp lands ~0.04% low end-to-end on your own test frame and ~0.8% low at `qk = 0.001`.
- `pk = 1` is the other NaN trigger — a bin where every treated unit converted returns NaN on master too — and the new test covers only `pk = 0`.

PR #1000, opened 2026-08-11
Take the KL limits at pk = 0 and pk = 1 instead of clamping

arpitjain099:
Follow-up to your review on #999. Both points were right, so this takes the exact branches.

`_kl_divergence` in `causalml/inference/tree/_uplift/_criterion.pyx` (lines 99-104) already does it:

```
if pk == 0.0:
    s = -log(1.0 - qk)
elif pk == 1.0:
    s = -log(qk)
else:
    s = pk * log(pk / qk) + (1.0 - pk) * log((1.0 - pk) / (1.0 - qk))
```

The clamp is the same amount of code and lands low. Measured against the limits:

```
pk=0.0 qk=0.5  : clamp=0.6931323650  exact=0.6931471806  -0.0021%
pk=0.0 qk=0.05 : clamp=0.0512814233  exact=0.0512932944  -0.0231%
pk=0.0 qk=0.001: clamp=0.0009925916  exact=0.0010005003  -0.7905%
```

So the error grows as the control conversion rate falls, which is exactly the regime where the bins with no converters show up in the first place.

On `pk = 1`: right, the test in #999 only reached `pk = 0`. The production path to `pk = 1` is a bin in which everyone converted in both arms, since `_GetNodeSummary` smooths a missing count to 1 and that only leaves `y_mean` at exactly 1 when the bin has no zeros at all. `test_filter_kl_bin_where_everyone_converted` drives that through `filter_D`.

I did not copy the `if qk == 0.0: return 0.0` early return that sits above the branches in the Cython version. It would change scores that the current clamp already produces, and a control rate of 0 against a non-zero treatment rate is a real divergence rather than none. Happy to align that too if you would rather the two match exactly.

Verification: `test_kl_divergence_at_the_limits` is parametrized over five (pk, qk) pairs and compares against the closed form at `rel=1e-12`. All five fail on master and pass with this change. `test_filter_kl_bin_where_everyone_converted` passes either way, since the clamp from #999 already keeps it finite; it is there to hold the `pk = 1` branch. Run with `--noconftest`, since the fixtures in `tests/conftest.py` need the compiled tree extensions. `black --check` is clean.

### DIFF
--- a/causalml/feature_selection/filters.py
+++ b/causalml/feature_selection/filters.py
@@ -377,6 +377,15 @@ class FilterSelect:
             pk (float): Probability of class 1 in treatment group
             qk (float): Probability of class 1 in control group
         """
+        # Identical arms have no divergence. This is checked before the clamp
+        # below, which would otherwise move qk away from pk and report a
+        # difference between two arms that behaved the same. It is also the only
+        # way pk reaches 0 or 1 through filter_D: _GetNodeSummary smooths a
+        # missing count to 1, so a degenerate pk means the whole bin is
+        # degenerate and qk holds the same value.
+        if pk == qk:
+            return 0.0
+
         eps = 0.1**6
         if qk < eps:
             qk = eps
--- a/causalml/feature_selection/filters.py
+++ b/causalml/feature_selection/filters.py
@@ -382,11 +382,16 @@ class FilterSelect:
             qk = eps
         elif qk > 1 - eps:
             qk = 1 - eps
-        if pk < eps:
-            pk = eps
-        elif pk > 1 - eps:
-            pk = 1 - eps
-        S = pk * np.log(pk / qk) + (1 - pk) * np.log((1 - pk) / (1 - qk))
+
+        # pk = 0 and pk = 1 send one of the two terms to 0 * log(0), which numpy
+        # evaluates as NaN rather than the 0 the limit gives. Take the limits
+        # directly, as _kl_divergence in inference/tree/_uplift/_criterion.pyx does.
+        if pk == 0:
+            S = -np.log(1 - qk)
+        elif pk == 1:
+            S = -np.log(qk)
+        else:
+            S = pk * np.log(pk / qk) + (1 - pk) * np.log((1 - pk) / (1 - qk))
         return S
 
     def _evaluate_KL(self, nodeSummary, control_group="control"):
--- a/tests/test_feature_selection.py
+++ b/tests/test_feature_selection.py
@@ -130,3 +130,44 @@ def test_filter_kl_bin_with_no_conversions():
 
     assert np.isfinite(imp["score"].values[0])
     assert imp["rank"].values[0] == 1
+
+
+@pytest.mark.parametrize(
+    "pk, qk",
+    [(0.0, 0.5), (1.0, 0.5), (0.0, 0.001), (1.0, 0.001), (0.0, 0.999)],
+)
+def test_kl_divergence_at_the_limits(pk, qk):
+    """``pk`` at 0 or 1 must give the limit, not an epsilon-shifted approximation.
+
+    Clamping ``pk`` keeps the result finite but biased low, by ~0.8% at
+    ``qk = 0.001``. ``_kl_divergence`` in
+    ``causalml/inference/tree/_uplift/_criterion.pyx`` already takes the limits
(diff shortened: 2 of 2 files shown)


jeongyoonlee (commented):
**Blocking**

- `qk` is still clamped before the new branches, so `pk = qk = 0` returns `-log(1 - 1e-6)` = 1e-6 where master returns exactly 0 — the frame in `test_filter_kl_bin_where_everyone_converted` scores `0.0` on master and `8.0e-07` here, a divergence between two identical arms.
- `_GetNodeSummary` smooths a missing count to 1, so `pk` reaches exactly 0 or 1 only when both arms in the bin are degenerate, which puts `qk` at the same value; over 3000 random frames, 3358 calls reached `pk` in {0, 1} and none had `qk` off {0, 1}.
- `if pk == qk: return 0.0` above the clamp restores both frames to the master values and keeps the limits intact — `(0.0, 0.5)` gives 0.6931471806 and `(0.0, 0.001)` gives 0.0010005003.

**Non-blocking**

- The five pairs in `test_kl_divergence_at_the_limits` are all degenerate `pk` with non-degenerate `qk`, none of which `_GetNodeSummary` can emit; `(0.0, 0.0)` and `(1.0, 1.0)` are the reachable ones.
- `test_filter_kl_bin_where_everyone_converted` asserts only finiteness and rank, which both hold on master; `assert imp["score"].values[0] == 0` would pin the identical-arms case.
- Leaving out `if qk == 0.0: return 0.0` is right, since it returns 0 for `qk = 0` with `pk > 0`; the `pk == qk` guard covers `pk = qk = 0` in both files and also `pk = qk = 1`, which the Cython version returns `EPS` for.

arpitjain099:
You are right on all three, and the blocking one is a regression I introduced. Fixed with the guard you suggested.

Reproduced it first. `pk = qk = 0` on master versus this branch as it stood:

```
pk=0.0 qk=0.0: master=0.0000000000e+00  branch=1.0000005000e-06
pk=1.0 qk=1.0: master=0.0000000000e+00  branch=1.0000005000e-06
```

and end to end the frame in `test_filter_kl_bin_where_everyone_converted` scored `8.000004000232713e-07`, matching your number. Clamping `qk` moved it away from `pk`, and then the `pk == 0` branch reported a divergence between two arms that behaved identically. Worse than the bias I was trying to remove.

I also checked the reachability claim rather than take it on faith, since it is the part that decides whether the guard is sufficient. Instrumenting `_kl_divergence` and running 300 random frames through `filter_D`: 1906 calls, 92 with `pk` in {0, 1}, and every one of those had `qk` equal to `pk`. The only degenerate pairs that appeared were `(0.0, 0.0)` and `(1.0, 1.0)`. Same conclusion as your 3000-frame run.

`if pk == qk: return 0.0` now sits above the clamp. It is not only a degenerate-case patch: KL of a distribution against itself is 0 by definition, so it holds for any equal pair, and it leaves the limits alone:

```
pk=0.0 qk=0.0   -> 0.0000000000
pk=1.0 qk=1.0   -> 0.0000000000
pk=0.0 qk=0.5   -> 0.6931471806
pk=0.0 qk=0.001 -> 0.0010005003
```

Both of your non-blocking points are in too. `test_kl_divergence_identical_arms` covers `(0, 0)`, `(1, 1)` and a non-degenerate `0.3` pair, and its docstring records why those are the reachable ones; the existing limits test now says in its docstring that its pairs are direct-caller coverage that `filter_D` cannot emit. `test_filter_kl_bin_where_everyone_converted` asserts `score == 0` rather than just finiteness. Removing the guard fails 3 of those.

On the Cython version: agreed, and it has the same gap at `pk = qk = 1`, where `qk` clamps to `1 - EPS` and the `pk == 1` branch returns `EPS`. I have left `_criterion.pyx` alone here because that scalar feeds the normalization factor for every divergence criterion, so changing it shifts tree splits, which is not something to slip into a feature-selection fix. Happy to send it as its own PR if you want it.

One process note worth stating: I can now build the extensions locally, so this run is the real suite rather than the `--noconftest` subset I mentioned before. `tests/test_feature_selection.py` is 18 passed with the fixtures active, `tests/test_cate_scoring.py` alongside it is 45 passed, and `black --check` is clean.

jeongyoonlee (approved):
LGTM

jeongyoonlee (approved):
LGTM

PR #1001, opened 2026-08-12
Add honest estimation to causal trees and forests (#584)

### LINKED ISSUE #584: When add the Honest Tree to the CausalTreeRegressor?
Hi,@alexander-pv. Do you have any plans to add the Honesty Tree to the CausalTreeRegressor?

jeongyoonlee:
## Proposed changes

Adds honest estimation to `CausalTreeRegressor` and `CausalRandomForestRegressor`, on by default. Closes #584.

## What honest estimation is

A causal tree partitions the feature space into leaves and reports one treatment effect per leaf — the CATE, `tau(x) = E[Y(1) - Y(0) | X = x]`, held constant within a leaf.

When the same rows choose the splits and supply the leaf estimate, the estimate keeps the noise that motivated the split: the search keeps whichever split shows the largest treated-minus-control difference, and the maximum of many noisy differences is biased away from zero.

Honest estimation gives each row one job ([Athey & Imbens 2016](https://arxiv.org/abs/1504.01132)):

1. Split the sample into a **structure** half and an **estimation** half.
2. Grow the tree on the structure half.
3. Discard the leaf values and recompute each leaf's per-group mean outcome from the estimation half, structure fixed.

`honesty=True` is the new default. `estimation_sample_size` (default `0.5`) is the fraction held out for step 3, and the split is stratified on treatment. This matches `grf`'s [`honesty = TRUE`](https://grf-labs.github.io/grf/reference/causal_forest.html) and EconML's [`honest=True`](https://econml.azurewebsites.net/spec/estimation/forest.html).

**This changes fitted models and predictions.** An existing script gets a different tree and different CATE estimates with no edit. `honesty=False` restores the previous behavior and is bit-identical to master.

Data with no treatment effect anywhere, where every leaf should report 0 — 200 replications, structure held fixed across both arms:

| leaves estimated | mean absolute `tau_hat` |
| --- | --- |
| in-sample (`honesty=False`) | 0.163 |
| honest (`honesty=True`) | 0.095 |

The cost is variance: each half sees half the rows, so the tree is shallower and each leaf noisier. Use `honesty=False` when halving the sample leaves too little to split on.

### Implementation notes

A causal-tree leaf stores one mean outcome **per treatment group** and `predict` differences them, so the re-estimation rewrites those means in `tree_.value`. It is weighted, unlike `UpliftTreeClassifier`'s: ignoring `sample_weight` leaves an ATE error four times larger once inverse-propensity weights are in play (0.285 vs 0.073). A group with no estimation rows in a leaf keeps its structure-half value.

On the forest, honesty composes with `bootstrap` rather than replacing it: the bootstrap counts arrive as each tree's `sample_weight` and weight both the structure fit and the re-estimation. `bootstrap=False` gives the honest split alone.

## Parity with grf and EconML

`docs/examples/causal_tree_honesty_parity.ipynb` (new) runs CausalML, EconML 0.17.0 and `grf` 2.6.1 on the same [`synthetic_data`](https://causalml.readthedocs.io/en/latest/causalml.html#module-causalml.dataset) draws — the designs from [Nie & Wager 2018](https://arxiv.org/abs/1712.04912) — scored on held-out rows against the known `tau`. `grf` runs through an `Rscript` subprocess and the cells skip themselves when R is absent.

Randomized trial, pairwise correlation between the three libraries' honest forests (agreement, not accuracy):

| | CausalML | EconML | grf |
| --- | --- | --- | --- |
| **CausalML** | 1.000 | 0.922 | 0.970 |
| **EconML** | 0.922 | 1.000 | 0.938 |
| **grf** | 0.970 | 0.938 | 1.000 |

True null, 25 replications, mean absolute `tau_hat` where 0 is correct:

| | in-sample | honest | reduction |
| --- | --- | --- | --- |
| CausalML | 0.2397 | 0.0628 | 74% |
| EconML | 0.1352 | 0.0716 | 47% |

### Confounded design

On `mode=1`, where the treatment probability `e(X)` depends on the features, CausalML's correlation with the true `tau` is lower than both references. Regenerating the identical design with treatment assigned at random, changing nothing but `e(X)`:

| design | CausalML | EconML |
| --- | --- | --- |
| confounded | 0.371 | 0.776 |
| randomized | 0.951 | 0.845 |

Confounding accounts for the gap. Every CausalML causal criterion compares raw group means, which estimate `tau` only under random assignment; `grf` and EconML relabel each node using treatment residualized against the node mean.

`sample_weight` reaches the split search, the leaf means and the re-estimation, so inverse-propensity weights — `1/e(X_i)` if treated, `1/(1 - e(X_i))` if control — correct it (10 seeds, cross-fitted propensity):

| | corr w/ true tau | ATE error | CATE RMSE |
| --- | --- | --- | --- |
| CausalML, as-is | 0.381 | 0.389 | 0.433 |
| CausalML + IPW `sample_weight` | 0.844 | 0.067 | 0.142 |
| EconML `CausalForest` | 0.788 | 0.269 | 0.301 |

This was undocumented, and it is the correction Athey & Imbens give in section 7: propensity score weighting, renormalized within each leaf. Both classes now carry a note with the recipe. Use plain IPW rather than stabilized or overlap weights — plain balances the weighted group sizes exactly and `groups_penalty=0.5` penalizes that imbalance.

## `ccp_alpha="cv"`, off by default

`honesty=True` is one of three pieces of the paper's CT-H procedure. `ccp_alpha="cv"` adds the other two, both from section 3.2:

- **The variance penalty scaled by `1 + N_structure/N_estimation`** (a factor of 2 at an even split). The objective rewards spread between leaves and penalizes within-leaf variance, since a high-variance leaf will not reproduce when re-estimated on other rows.
- **Tree size chosen by cross-validation.** [Cost-complexity pruning](https://scikit-learn.org/stable/modules/tree.html#minimal-cost-complexity-pruning) gives a nested sequence of subtrees indexed by a per-leaf penalty `alpha`; `cv_folds`-fold CV scores each with the same honest objective on the held-out fold.

Without it the tree grows until `min_samples_leaf` / `min_group_samples` stop it, and the variance penalty only *ranks* splits rather than *choosing tree size*, which is its job in the paper.

It lives on `ccp_alpha` because that is the parameter it chooses; a separate `cv` flag was rejected because sklearn reads `cv=None` as "use the default CV strategy" rather than "off". It requires `honesty=True` and raises otherwise. `get_params()["ccp_alpha"]` still returns `"cv"` after fitting; the resolved float is on `ccp_alpha_`.

The scaling alone measures nothing (paired over 24 seeds, p = 0.40 and 0.81). The CV sizing is what moves the numbers — held-out CATE RMSE, single tree, 16 seeds, paired:

| noise | `honesty` only | + `ccp_alpha="cv"` | change | leaves |
| --- | --- | --- | --- | --- |
| σ=0.1 | 0.1212 | 0.1192 | −1.7% (n.s.) | 46 → 14 |
| σ=0.5 | 0.2128 | 0.1676 | −21.2% | 46 → 13 |
| σ=1.0 | 0.3499 | 0.2287 | −34.6% | 46 → 10 |
| σ=2.0 | 0.7011 | 0.2977 | −57.5% | 46 → 8 |

Better in 16 of 16 seeds above σ=0.1, p < 0.0001; at σ=0.1 there is no overfitting to remove and it keeps the deeper tree. Cost is `cv_folds` extra fits.

This applies to a single tree. On a forest it measured no gain, paired over 10 seeds:

| σ | estimator | off | on | change | on better |
| --- | --- | --- | --- | --- | --- |
| 0.5 | single tree | 0.2125 | 0.1593 | −25.1% | 10/10 |
| 0.5 | forest (100) | 0.0842 | 0.1304 | +54.9% | 0/10 |
| 2.0 | single tree | 0.7136 | 0.3223 | −54.8% | 10/10 |
| 2.0 | forest (100) | 0.1836 | 0.1999 | +8.9% | 2/10 |

Averaging across trees already removes the variance the pruning targets. Both docstrings and the changelog carry these numbers.

## Implementation cross-check

Read against `grf` ([`TreeTrainer.cpp`](https://github.com/grf-labs/grf/blob/master/core/src/tree/TreeTrainer.cpp), `Tree.cpp`, [`causal_forest.R`](https://github.com/grf-labs/grf/blob/master/r-package/grf/R/causal_forest.R)) and EconML ([`_tree_classes.py`](https://github.com/py-why/EconML/blob/main/econml/tree/_tree_classes.py)). Agreements: 50/50 default split; the structure half is smaller than a non-honest fit's sample in `grf` too, rather than compensated for; weights apply within each half.

Two deliberate divergences, both in the `honesty` docstring:

| | grf | EconML | here |
| --- | --- | --- | --- |
| split | unstratified subsample | unstratified `shuffle` + halve | stratified on treatment |
| leaf empty on the estimation half | pruned into its sibling (`honesty.prune.leaves=TRUE`) | prevented at split time by a second `criterion_val` | keeps its structure-half value |

CausalML stratifies because a single tree is a first-class object here rather than only a forest component, and because the builder enforces `min_group_samples`. That makes the empty-leaf case rare: across 200 honest fits in 5 configurations, no leaf lost a treatment arm on the estimation half.

Node-level residualization is not part of this algorithm. [Athey & Imbens 2016](https://arxiv.org/abs/1504.01132) and [Wager & Athey 2018](https://arxiv.org/abs/1510.04342) have none; it arrives with [Generalized Random Forests 2019](https://arxiv.org/abs/1610.01271), which EconML's `grf.CausalForest` implements and which describes Wager & Athey 2018 as "almost equivalent to a generalized random forest without centering".

## Where the code lives

| | |
| --- | --- |
| `causal/causaltree.py` | `_fit_honest` (the split), `_honest_reestimate` (rewrites `tree_.value`), `_select_ccp_alpha` / `_make_fold_tree` (the CV) |
| `causal/_criterion.pyx` | `CausalMSE.node_impurity` / `children_impurity` — the objective and the `1 + train_to_est_ratio` scaling |
| `causal/causalforest.py` | the same four parameters, forwarded per tree via `estimator_params` |
| `tests/test_causal_trees.py` | 20 new tests, including the `honesty=False` bit-identity check |

Editing a `.pyx`/`.pxd` needs `uv pip install -e . --no-deps`.

## Types of changes

## Checklist

## Further comments

`TestCausalTreeCase.prepare_model` is pinned to `honesty=False`. `test_ate` asserts `ape(...) < 0.5` on a fixture where both modes sit near that line: at `n_treatments=2` the measured APE is 0.405 in-sample and 0.544 honest, and across five other seeds the in-sample mean is 0.515. The reason is in a comment at the call site.

Follow-ups, each worth its own issue:

- Built-in propensity support: a `p` argument that forms the IPW weights, so callers do not hand-roll them.
- Honest `node_pvalues`. It remains an in-sample t-test over the full `fit` sample under `honesty=True`, and the docstring says so; computing it on the estimation half would make it valid for post-selection inference.
- `autosectionlabel_prefix_document = True` in `docs/conf.py`. The changelog's repeated subsection headings produce six duplicate-label warnings, one of them added here; the setting clears all six but can invalidate existing `:ref:` targets.


ras44 (approved):
nice!

PR #1004, opened 2026-08-12
Let UpliftTreeClassifier prune inside fit (#1003)

### LINKED ISSUE #1003: Make UpliftTreeClassifier pruning reachable from fit()

`UpliftTreeClassifier.prune()` improves held-out qini substantially, but reaching it requires the caller to hold out validation rows themselves, fit, and then call `prune()` as a second step. Nothing in `fit` suggests it exists, and `UpliftRandomForestClassifier` never calls it.

Measured on `make_uplift_classification` (n=3000, 6 seeds, 30% of the training rows held out for pruning, scored on a separate 40% test split):

| `max_depth` | leaves before → after | qini unpruned | qini pruned | change |
| --- | --- | --- | --- | --- |
| 3 (default) | 7.7 → 4.2 | 0.579 | 0.839 | +0.26 |
| `None` | 82.7 → 54.8 | 0.253 | 0.394 | +0.14 |

Caveat on the numbers: one simulated design, and the unpruned arm was fitted on 30% less data than a plain `fit` would use, so the magnitudes are not a benchmark. The direction is consistent across seeds and is larger at the default depth than at full depth.

Do the split-and-prune inside `fit`, off by default:

```python
UpliftTreeClassifier(prune_fraction=0.3, min_gain=0.0001)
```

`prune_fraction=None` (default) keeps today's behavior. When set, `fit` holds out that fraction stratified on (treatment, y), grows the tree on the rest, and prunes on the holdout with the existing `prune()`. `prune()` stays as it is for callers who manage their own split.

Then `min_gain="cv"`: select the threshold by `cv_folds`-fold cross-validation rather than fixing it, each fold growing on k-1 folds and pruning on the held-out one.

`ccp_alpha="cv"`, which #1001 adds to `CausalTreeRegressor`, does not port. Cost-complexity pruning needs `R(t) >= R(T_t)`, and the uplift criteria are gain-based: `node_impurity` is a positive constant by design so the builder's `impurity <= EPSILON` early-leaf never fires. On a fitted uplift tree `ccp_pruning_path` returns 28 alphas of which 26 are negative, leaving `{0.0, 0.997}` after the non-negative filter — keep the whole tree, or prune to the root. `prune()`'s `minGain` is the right knob for a gain-based criterion.

**Additional context**

`UpliftRandomForestClassifier` does not prune at all and is the more commonly used class. Worth measuring separately before extending this to it: the equivalent option on `CausalRandomForestRegressor` measured *worse* than leaving it off (#1001), since averaging across trees already removes the variance pruning targets.

jeongyoonlee:
## Proposed changes

`prune_fraction` (default `None`, off) makes `UpliftTreeClassifier.fit` do the split-and-prune itself. Closes #1003.

## Why prune an uplift tree

An uplift tree models the change in response caused by a treatment, so it chooses splits by how much they separate the treated arm from the control arm: `KL`, `ED` and `Chi` by the gain in divergence between the two arms' outcome distributions, `DDP` by how much the treated-minus-control response rate differs between the children ([Rzepakowski & Jaroszewicz 2012](https://doi.org/10.1007/s10115-011-0434-0); [full list](https://causalml.readthedocs.io/en/latest/methodology.html#uplift-tree)).

Each is a difference of two estimated rates, whose variances add, so an uplift split is noisier than a classification split and a deep tree overfits readily. Validation pruning is the standard answer: grow the tree, then walk it bottom-up on rows it never saw and collapse any subtree that does not gain `min_gain` on those rows.

`prune()` already implemented this. What was missing was a way to reach it from `fit` — you had to split the data yourself, fit, then call `prune()`, with nothing in `fit` to suggest it existed.

`prune_fraction=0.3` holds out 30% of the rows stratified on (treatment, outcome), grows the tree on the rest, and runs `prune()` on the holdout with `min_gain` / `prune_rule`. `prune()` is unchanged for callers who manage their own holdout, and `n_nodes_before_pruning_` records the size pruning started from.

### Ordering, which matters twice

- **The pruning rows come out before the honest split**, so neither the split search nor the estimation half sees them. A split kept because it helps on the rows that chose it is what pruning exists to catch.
- **Pruning runs inside the ordinary fit path**, between growing the tree and estimating its leaves, so a node promoted to a leaf is re-estimated like any other rather than keeping its internal-node value.

Nothing is retained on the estimator to make that work: `UpliftRandomForestClassifier` fits each tree on its own full-size bootstrap copy of `X`, so an array parked on a fitted tree is multiplied by `n_estimators`. A test asserts a fitted tree holds no row-length arrays.

### Measured

`make_uplift_classification` (n=3000, 6 seeds, `max_depth=None`, `min_samples_leaf=20`), held-out qini — area between the cumulative-uplift curve and the random-targeting line, so negative is worse than random:

| | leaves | qini |
| --- | --- | --- |
| off (default) | 115.0 | -1.74 |
| `prune_fraction=0.3` | 53.8 | -1.44 |
| `prune_fraction=0.3, honesty=True` | 35.7 | 0.83 |

An unpruned tree at that depth is badly overfit. One simulated design, so treat the magnitudes as indicative; the direction is consistent across seeds.

## Bug fix included

`prune()` replaced `tree_` without rebuilding `_node_group_counts` — per-node treated/control row counts, indexed by node id, read by the uplift-score p-value and the plot's `group_size`.

Pruning renumbers nodes, so the ids no longer line up. Because the stale array is *longer* than the pruned tree, the lookup did not raise; it returned another node's counts. On a 283-node tree pruned to 183, the array stayed at 283 rows.

`prune()` now carries the counts over to the pruned ids, so a standalone call is fixed as well as the fit-time path. A surviving node is reached by the same rows and collapsing a subtree does not change its root's count, so the values transfer unchanged and only the ids move. The remap replays `_build_pruned_tree`'s traversal — stack-based preorder, right child pushed before left — and matches a full recomputation exactly across seeds.

## Not done here

`ccp_alpha="cv"`, which #1001 added to `CausalTreeRegressor`, does not port.

[Cost-complexity pruning](https://scikit-learn.org/stable/modules/tree.html#minimal-cost-complexity-pruning) ranks subtrees by `alpha = (R(t) - R(T_t)) / (|leaves(T_t)| - 1)`, the price per leaf of collapsing `T_t` into `t`. That ordering needs `R(t) >= R(T_t)`, which holds for squared error or Gini because a split cannot increase them.

The uplift criteria are gain-based instead — `node_impurity` is a positive constant by design, so the builder's `impurity <= EPSILON` early-leaf never fires — and the inequality does not hold. `ccp_pruning_path` on a fitted uplift tree returns 28 alphas of which 26 are negative, leaving `{0.0, 0.997}` after the non-negative filter: the whole tree, or the root.

CV-selected `min_gain` is the natural next step and is left for its own change. `UpliftRandomForestClassifier` still does not prune; the analogous option on `CausalRandomForestRegressor` measured worse than leaving it off (#1001), so that wants measuring before it is extended.

## Where the code lives

| | |
| --- | --- |
| `_uplift/uplifttree.py` | `_maybe_prune` (called from `fit`), `prune` (unchanged entry point), `_remap_node_group_counts` (the bug fix) |
| `tests/test_uplift_trees_kernel.py` | 6 new tests, including a call-order check that pruning precedes the honest re-estimation, and one asserting a fitted tree retains no training rows |

## Types of changes

## Checklist

### DIFF
--- a/causalml/inference/tree/_uplift/uplifttree.py
+++ b/causalml/inference/tree/_uplift/uplifttree.py
@@ -28,6 +28,36 @@ from ._tree import BaseUpliftDecisionTree
 from .._tree._tree import Tree, build_pruned_tree_from_mask
 
 
+def _check_fraction(name: str, value, allow_none: bool = False) -> None:
+    """Reject a held-out fraction that is not strictly inside (0, 1).
+
+    Both fractions reach ``train_test_split`` as ``test_size``, which rejects
+    out-of-range values but reports them against its own parameter name rather
+    than the one the caller passed. ``0.0`` needs catching here in particular:
+    it is a plausible way to write "hold nothing out", and nothing downstream
+    would have complained about it.
+
+    Args:
+        name (str): parameter name, for the message
+        value: the value to check
+        allow_none (bool): whether ``None`` is a valid value (off)
+
+    Raises:
+        ValueError: if the value is not a float strictly between 0 and 1
+    """
+    if allow_none and value is None:
+        return
+    try:
+        fraction = float(value)
+    except (TypeError, ValueError):
+        fraction = float("nan")
+    if not 0.0 < fraction < 1.0:
+        allowed = "a float strictly between 0 and 1"
+        if allow_none:
+            allowed += " or None"
+        raise ValueError(f"{name} must be {allowed}, got {value!r}")
+
+
 class _UpliftTreeNode:
     """A minimal ``DecisionTree``-shaped node for ``plot.py``.
 
@@ -131,7 +161,12 @@ class _KernelUpliftTreeClassifier(SerializableLearner, BaseUpliftDecisionTree):
         Returns:
             self
         """
-        if self.prune_fraction:
+        # Validated here rather than in __init__ so the constructor stores its
+        # arguments verbatim (sklearn get_params / clone round-trip).
+        _check_fraction("estimation_sample_size", self.estimation_sample_size)
+        _check_fraction("prune_fraction", self.prune_fraction, allow_none=True)
+
+        if self.prune_fraction is not None:
             return self._fit_with_pruning(
                 X=X,
                 treatment=treatment,
@@ -641,7 +676,9 @@ class UpliftTreeClassifier(_KernelUpliftTreeClassifier):
     ``min_gain`` / ``prune_rule``. The pruning rows are taken out before the honest
     split, so neither the split search nor the estimation half sees them, and
     ``n_nodes_before_pruning_`` records the size pruning started from. :meth:`prune`
-    is unchanged for callers managing their own holdout.
+    is unchanged for callers managing their own holdout. ``prune_fraction`` and
+    ``estimation_sample_size`` are held-out fractions, so ``fit`` rejects anything
+    outside ``(0, 1)`` — including ``0.0``, which would otherwise read as off.
 
     On ``make_uplift_classification`` (n=3000, 6 seeds, ``max_depth=None``,
     ``min_samples_leaf=20``), ``prune_fraction=0.3`` took held-out qini from -1.74 to
--- a/causalml/inference/tree/_uplift/uplifttree.py
+++ b/causalml/inference/tree/_uplift/uplifttree.py
@@ -150,10 +150,8 @@ class _KernelUpliftTreeClassifier(SerializableLearner, BaseUpliftDecisionTree):
             super().fit(
                 X=X_enc, y=y_2dim, sample_weight=sample_weight, check_input=check_input
             )
+            self._maybe_prune()
             self._node_group_counts = self._compute_node_group_counts(X_enc, y_2dim)
-            # Recorded for `_fit_with_pruning`, which has to redo both against the
-            # pruned tree. Structure rows only; there is no estimation half here.
-            self._fit_splits = (X_enc, y_2dim, None, None)
             return self
 
         # Honest approach (Athey & Imbens 2016): grow the tree on one split and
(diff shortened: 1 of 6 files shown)


ras44 (commented):
might add a prune fraction validation tests in the constructor and a defensive test in _fit_with_pruning:

something like:

In **UpliftTreeClassifier.init** (after forwarding prune_fraction to super or after assigning it locally if you do so):

```
# validate prune_fraction passed at the public constructor
if prune_fraction is not None:
    try:
        pf = float(prune_fraction)
    except (TypeError, ValueError):
        raise ValueError(f"prune_fraction must be a number in (0, 1) or None; got {prune_fraction!r}")
    if not (0.0 < pf < 1.0):
        raise ValueError(f"prune_fraction must be strictly between 0 and 1; got {prune_fraction!r}")

```


**in _KernelUpliftTreeClassifier.init** 
```
# validate prune_fraction (None or 0 < prune_fraction < 1)
if self.prune_fraction is not None:
    try:
        pf = float(self.prune_fraction)
    except (TypeError, ValueError):
        raise ValueError(f"prune_fraction must be a number in (0, 1) or None; got {self.prune_fraction!r}")
    if not (0.0 < pf < 1.0):
        raise ValueError(f"prune_fraction must be strictly between 0 and 1; got {self.prune_fraction!r}")

```

jeongyoonlee:
Agreed on validating, with one change: it goes in `fit` rather than `__init__` — sklearn's contract is that a constructor only stores its arguments, which `get_params` / `clone` round-trip on. Same place `CausalTreeRegressor.fit` validates `ccp_alpha` (`causaltree.py:281`).

The case worth catching is `prune_fraction=0.0`: `fit` branched on truthiness, so it read as off and skipped pruning silently. Out-of-range and non-numeric values already raised, but from `train_test_split` and against `test_size`.

`ea4705a` rejects anything outside `(0, 1)` for `prune_fraction` (`None` still means off) and for `estimation_sample_size`, which had the same gap, with tests for both.

ras44 (approved):

PR #1030, opened 2026-08-18
Fix #1028: supply default LightGBM learner for cate scoring metrics

su-jin1425:
## Proposed changes

This PR fixes #1028 by providing a default `LGBMRegressor` when no outcome learner is supplied to the CATE scoring metrics.

Previously, `_resolve_outcome_learners` raised a `ValueError` when neither a learner nor both treatment/control outcome learners were provided. This prevented the metrics from being used with their expected default behavior.

The fix:
- Uses `LGBMRegressor` as the default outcome learner.
- Keeps the existing behavior when an explicit learner or both outcome learners are provided.
- Preserves the existing `ValueError` fallback when LightGBM is unavailable.
- Uses the existing LightGBM configuration: `num_leaves=64`, `learning_rate=0.05`, `n_estimators=300`.

## Before
before

## After
after

### Before
```powershell
git checkout 35f6e734a478eea6fdf1cfaf1b85e0f04ae99b83; python -c "import numpy as np,pandas as pd; from causalml.metrics import dr_score,plug_in_t_score; rng=np.random.default_rng(42); X=rng.normal(size=(100,3)); w=rng.integers(0,2,100); y=1+X[:,0]+0.5*X[:,1]+2*w+rng.normal(size=100); df=pd.DataFrame({'y':y,'w':w,'model':2+0.5*X[:,0]}); expected='Either `learner` or both `control_outcome_learner` and `treatment_outcome_learner` must be specified.'; print('BEFORE: dr_score'); assert (lambda: True)(); exec('try:\n dr_score(df,X=X,outcome_col=\"y\",treatment_col=\"w\",n_folds=2,random_state=42)\nexcept ValueError as e:\n assert str(e)==expected\n print(\"PASS: expected ValueError\")'); print('BEFORE: plug_in_t_score'); exec('try:\n plug_in_t_score(df,X=X,outcome_col=\"y\",treatment_col=\"w\",n_folds=2,random_state=42)\nexcept ValueError as e:\n assert str(e)==expected\n print(\"PASS: expected ValueError\")')"
```

### After
```powershell
[...]

PR #1031, opened 2026-08-18
Fix S-learner estimate_ate DataFrame handling and return shape (#1026)

### LINKED ISSUE #1026: S-learner estimate_ate crashes on DataFrame input and returns a different shape than the other meta-learners
Two related problems in `BaseSLearner.estimate_ate` (`causalml/inference/meta/slearner.py`). Both observed on master (`18ebfb0`).

**1. Crash on DataFrame input.** `StatsmodelsOLS.fit` stores `self.conf_ints = self.model.conf_int(alpha=...)` (`slearner.py:48`). With a DataFrame `X`, statsmodels returns that as a DataFrame, and `estimate_ate` indexes it like an ndarray — `self.models[group].conf_ints[0, 0]` (`slearner.py:449`) — raising `KeyError: (0, 0)`. The same call with `X.values` works.

```python
import pandas as pd
from causalml.dataset import fetch_lalonde
from causalml.inference.meta import LRSRegressor

data = fetch_lalonde()
X = pd.DataFrame(data.data, columns=data.feature_names)
LRSRegressor().estimate_ate(X=X, treatment=data.treatment, y=data.target)
# KeyError: (0, 0); passing X.values returns (1676, 350, 3003)
```

Suggestion: `np.asarray` the interval in `StatsmodelsOLS.fit`.

**2. Return signature differs from every other meta-learner.** `BaseSRegressor.estimate_ate` has a `return_ci` parameter and returns a bare ATE unless `return_ci=True`; `BaseTRegressor`, `BaseXRegressor`, `BaseRRegressor` and `BaseDRLearner` have no such parameter and always return `(ate, lb, ub)`. Code written against one learner breaks on another.

Suggestion: make the S-learner return the triple unconditionally like the others (a behavior change worth a changelog entry), or add `return_ci` everywhere with a consistent default.

Grouped into one issue because both fixes touch the same function.

su-jin1425:
## Proposed changes

Fixes #1026.

This PR fixes two issues in `BaseSLearner.estimate_ate`:

- Fixes a `KeyError: (0, 0)` when `LRSRegressor.estimate_ate()` is called with a pandas DataFrame.
- Makes `BaseSLearner.estimate_ate()` consistently return `(ate, lb, ub)`, matching the other meta-learners.

The statsmodels parameters and confidence intervals are converted to NumPy arrays before positional indexing.

Tests were updated to reflect the consistent S-learner return signature, and a regression test was added for DataFrame input.

## Before & After
before and after

## Test
```powershell
function T($c,$n) {
    git checkout --force $c | Out-Null
    Write-Host "`n========== $n =========="

    @'
import numpy as np
import pandas as pd
from causalml.inference.meta import LRSRegressor, BaseSRegressor
from sklearn.linear_model import LinearRegression

rng = np.random.RandomState(42)
X = pd.DataFrame(rng.normal(size=(200, 5)), columns=list("abcde"))
t = rng.binomial(1, 0.5, 200)
y = 2*t + X["a"].to_numpy() + 0.5*X["b"].to_numpy() + rng.normal(size=200)

try:
    result = LRSRegressor().estimate_ate(X=X, treatment=t, y=y)
    print("DataFrame: PASS ->", result)
except Exception as e:
    print("DataFrame: FAIL ->", type(e).__name__, e)

X = rng.normal(size=(200, 5))
y = 2*t + X[:, 0] + 0.5*X[:, 1] + rng.normal(size=200)

result = BaseSRegressor(learner=LinearRegression()).estimate_ate(
    X=X, treatment=t, y=y
)

print("S-learner:", result)
print("Return type:", type(result).__name__)
print("Return length:", len(result))
'@ | python
}

T "35f6e734a478eea6fdf1cfaf1b85e0f04ae99b83" "BEFORE #1026 — BUG"
T "ef98264f8d7b8cce18666464cabb66b7abc0e402" "AFTER #1026 — FIXED"
```

## Types of changes

## Checklist

## Further comments

The `return_ci` argument for `BaseSLearner.estimate_ate()` was removed so that S-learners follow the same `estimate_ate()` return convention as the other meta-learners.

The original DataFrame failure was reproduced against the parent commit and verified as fixed. CI passes on Python 3.11 and 3.12, and lint checks pass.

### DIFF
--- a/causalml/inference/meta/slearner.py
+++ b/causalml/inference/meta/slearner.py
@@ -44,8 +44,8 @@ class StatsmodelsOLS:
         # Append ones. The first column is for the treatment indicator.
         X = sm.add_constant(X, prepend=False, has_constant="add")
         self.model = sm.OLS(y, X).fit(cov_type=self.cov_type)
-        self.coefficients = self.model.params
-        self.conf_ints = self.model.conf_int(alpha=self.alpha)
+        self.coefficients = np.asarray(self.model.params)
+        self.conf_ints = np.asarray(self.model.conf_int(alpha=self.alpha))
         return self
 
     def predict(self, X):
@@ -233,7 +233,6 @@ class BaseSLearner(BaseLearner):
         treatment,
         y,
         p=None,
-        return_ci=False,
         bootstrap_ci=False,
         n_bootstraps=1000,
         bootstrap_size=10000,
@@ -245,7 +244,6 @@ class BaseSLearner(BaseLearner):
             X (np.matrix, np.array, pd.DataFrame, pl.DataFrame, or pl.LazyFrame): a feature matrix
             treatment (np.array, pd.Series, or pl.Series): a treatment vector
             y (np.array, pd.Series, or pl.Series): an outcome vector
-            return_ci (bool, optional): whether to return confidence intervals
             bootstrap_ci (bool): whether to return confidence intervals
             n_bootstraps (int): number of bootstrap iterations
             bootstrap_size (int): number of samples per bootstrap
@@ -296,9 +294,7 @@ class BaseSLearner(BaseLearner):
             ate_lb[i] = _ate_lb
             ate_ub[i] = _ate_ub
 
-        if not return_ci:
-            return ate
-        elif return_ci and not bootstrap_ci:
+        if not bootstrap_ci:
             return ate, ate_lb, ate_ub
         else:
             t_groups_global = self.t_groups
--- a/docs/examples/feature_interpretations_example.ipynb
+++ b/docs/examples/feature_interpretations_example.ipynb
@@ -156,7 +156,7 @@
     "# base_algo = LinearRegression()\n",
     "\n",
     "slearner = BaseSRegressor(base_algo, control_name='control')\n",
-    "slearner.estimate_ate(X=X, treatment=w_multi, y=y)"
+    "slearner.estimate_ate(X=X, treatment=w_multi, y=y)[0]"
    ]
   },
   {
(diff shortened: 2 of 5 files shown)


jeongyoonlee (changes requested):
`docs/examples/meta_learners_with_synthetic_data_multiple_treatment.ipynb` still passes the removed `return_ci` to `BaseSRegressor.estimate_ate()`, so six cells now raise `TypeError`. The S-learner ATE section appears twice in the notebook, so each of these is duplicated:

- `ate_s = learner_s.estimate_ate(..., return_ci=False, bootstrap_ci=False)` — lines 153 / 2430
- `ate_s, ate_s_lb, ate_s_ub = learner_s.estimate_ate(..., return_ci=True, bootstrap_ci=False)` — 215 / 2524
- `ate_s_b, ate_s_lb_b, ate_s_ub_b = learner_s.estimate_ate(..., return_ci=True, bootstrap_ci=True, ...)` — 280 / 2596

The last two only need `return_ci` dropped. The first is a point-estimate-only demo with no equivalent after this change — the following cell displays `ate_s`, which is now a 3-tuple — so please either drop that cell pair or index the ATE out. Re-run the notebook afterwards so the stored outputs match.

Minor: `docs/examples/feature_interpretations_example.ipynb` ends a cell with `slearner.estimate_ate(X=X, treatment=w_multi, y=y)` on a `BaseSRegressor`; its stored output is now stale (array → 3-tuple).

su-jin1425:
I’ve updated both affected notebooks to remove the deprecated return_ci usage and handle the new (ate, lb, ub) return consistently. The changes are in commit 4cae40d. Full notebook execution wasn’t possible locally due to the missing Cython _criterion extension, but the affected cells were updated surgically without unrelated output changes.

su-jin1425:
Use https://github.com/uber/causalml/pull/1030 → https://github.com/uber/causalml/pull/1033 → https://github.com/uber/causalml/pull/1032 → https://github.com/uber/causalml/pull/1031 this order to avoid merge conflicts

jeongyoonlee (approved):
LGTM. Thanks!

PR #1032, opened 2026-08-19
Fix #1029: Thread random_state to BaseRLearner bootstraps

### LINKED ISSUE #1029: Metric bootstraps and R-learner cross-validation draw from global NumPy state
Several components draw randomness from the global NumPy state when no seed is passed, so identical code produces different numbers depending on what ran before it in the same process. Two measured instances, both on master (`18ebfb0`):

- `qini_score(..., return_ci=True)` with `random_state=None` returned a bootstrap SE of 47.0 in one run and 1.5 in another on identical data — same seed at program start, different preceding computations. The same applies to the bootstrap in `auuc_score`, `rate_score`, `dr_score` and `plug_in_t_score`.
- `BaseRRegressor`'s internal cross-validation is seeded from global state: with `XGBRegressor(random_state=42)` as the base learner and identical data, the fitted model's held-out RATE was 0.283 in one execution order and 0.612 in another.

The metric functions already accept `random_state`, and the docs now advise passing it, so the metric half is a documentation-plus-defaults question. The estimator half has no equivalent: `BaseRLearner` (and any learner using internal cross-fitting) exposes no seed parameter to propagate to its CV splitter.

Suggestion: accept a `random_state` on learners with internal data splits and thread it through, so a fully seeded pipeline is reproducible regardless of execution order. Independent of #1026-#1028: the estimator change lives in `causalml/inference/meta/rlearner.py`, and the metric-default change (if any) is separable from #1028's learner-resolution path.

su-jin1425:
## Proposed changes

Fixes #1029.

`BaseRLearner` bootstrap confidence intervals could depend on unrelated global NumPy RNG state even when `random_state` was provided.

This change:
- Creates a local NumPy RNG from `BaseRLearner.random_state`.
- Passes the RNG to the bootstrap sampling in `fit_predict()` and `estimate_ate()`.
- Adds a regression test covering both public bootstrap paths and verifying reproducibility after unrelated global NumPy RNG consumption.

The metric functions are unchanged because they already use their provided `random_state` for bootstrap sampling.

## Test

```powershell
$S="$env:TEMP\causalml-1029-review\test_1029.py";$B="$env:TEMP\causalml-1029-review\before";$A="$env:TEMP\causalml-1029-review\after";@'
import sys
import numpy as np
from sklearn.linear_model import LinearRegression
sys.path.insert(0, sys.argv[1])
from causalml.inference.meta.rlearner import BaseRRegressor

rng=np.random.RandomState(123)
X=rng.randn(160,4)
t=rng.binomial(1,.5,160)
y=1.5*X[:,0]-.8*X[:,1]+2*t+rng.randn(160)*.5

def run(noise):
    np.random.seed(2026)
    np.random.rand(noise)
    m=BaseRRegressor(learner=LinearRegression(),n_fold=3,random_state=42)
    _,lo,hi=m.fit_predict(X,y=y,treatment=t,return_ci=True,n_bootstraps=20,bootstrap_size=120,verbose=False)
    return lo,hi

a=run(0)
b=run(1000)
print("same lower:",np.array_equal(a[0],b[0]))
print("same upper:",np.array_equal(a[1],b[1]))
print("RESULT:","PASS" if np.array_equal(a[0],b[0]) and np.array_equal(a[1],b[1]) else "FAIL")
'@|Set-Content $S -Encoding UTF8;"BEFORE 35f6e734";python $S $B;"AFTER 52fd2734";python $S $A
```

## Types of changes

## Checklist

## Further comments

The regression test verifies that identical seeded `BaseRRegressor` runs produce identical confidence intervals regardless of unrelated global NumPy RNG consumption.

### DIFF
--- a/causalml/inference/meta/rlearner.py
+++ b/causalml/inference/meta/rlearner.py
@@ -4,6 +4,7 @@ import numpy as np
 from tqdm import tqdm
 from scipy.stats import norm
 from sklearn.model_selection import cross_val_predict, KFold, train_test_split
+from sklearn.utils import check_random_state
 
 try:
     from xgboost import XGBRegressor, XGBClassifier
@@ -312,12 +313,15 @@ class BaseRLearner(BaseLearner):
             )
 
             logger.info("Bootstrap Confidence Intervals")
+            rng = check_random_state(self.random_state)
             for i in tqdm(range(n_bootstraps)):
                 if p is None:
                     p = self.propensity
                 else:
                     p = self._format_p(p, self.t_groups)
-                te_b = self.bootstrap(X, treatment_np, y_np, p, size=bootstrap_size)
+                te_b = self.bootstrap(
+                    X, treatment_np, y_np, p, size=bootstrap_size, rng=rng
+                )
                 te_bootstraps[:, :, i] = te_b
 
             te_lower = np.percentile(te_bootstraps, (self.ate_alpha / 2) * 100, axis=2)
@@ -414,12 +418,15 @@ class BaseRLearner(BaseLearner):
             logger.info("Bootstrap Confidence Intervals for ATE")
             ate_bootstraps = np.zeros(shape=(self.t_groups.shape[0], n_bootstraps))
 
+            rng = check_random_state(self.random_state)
             for n in tqdm(range(n_bootstraps)):
                 if p is None:
                     p = self.propensity
                 else:
                     p = self._format_p(p, self.t_groups)
-                cate_b = self.bootstrap(X, treatment_np, y_np, p, size=bootstrap_size)
+                cate_b = self.bootstrap(
+                    X, treatment_np, y_np, p, size=bootstrap_size, rng=rng
+                )
                 ate_bootstraps[:, n] = cate_b.mean(axis=0)
 
             ate_lower = np.percentile(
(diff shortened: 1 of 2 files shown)


su-jin1425:
Use https://github.com/uber/causalml/pull/1030 → https://github.com/uber/causalml/pull/1033 → https://github.com/uber/causalml/pull/1032 → https://github.com/uber/causalml/pull/1031 this order to avoid merge conflicts

jeongyoonlee (changes requested):
`np.random.default_rng(self.random_state)` (`rlearner.py:315`, `:420`) rejects a `RandomState` instance, which `rlearner.py:62` documents as supported:

```python
BaseRRegressor(learner=LinearRegression(), random_state=np.random.RandomState(42)) \
    .fit_predict(X, y=y, treatment=t, return_ci=True)
# TypeError: SeedSequence expects int or sequence of ints for entropy not RandomState(MT19937)
```

Same for `estimate_ate(..., bootstrap_ci=True)`. Both work on master, so this is a regression.

Please use `sklearn.utils.check_random_state(self.random_state)` instead — it normalizes `None`/int/`RandomState` to a `RandomState`, which has `.choice()`, so the rest of each hunk is unchanged. That is also what the tree side already does (`match.py:143`, `causalforest.py:482`, `upliftforest.py:66`), and it matches sklearn's guidance: [Controlling randomness](https://scikit-learn.org/stable/common_pitfalls.html) recommends passing `RandomState` instances to estimators, and sklearn does not accept numpy `Generator` at all.

The new test only passes an int, so the instance path stays uncovered — please add it.

su-jin1425:
Done.

su-jin1425:
Done. 
CI / CD failed so i worked on it.

jeongyoonlee (approved):
LGTM. Thanks!

PR #1038, opened 2026-08-23
Cap default LGBM `num_leaves` to fold/arm sample size in `cate_scoring`

### LINKED ISSUE #1037: cate_scoring default learner is oversized for its own per-fold, per-arm call pattern
Low priority.

The default `LGBMRegressor(num_leaves=64, ...)` merged in #1030 (`cate_scoring.py:63`, `:275`, `:399`) is inherited from `visualize.py`'s TMLE helpers (`:343`, `:453`, `:696`, `:740`), which fit on the full frame. `dr_score` / `plug_in_t_score` fit per-fold **and** per-arm — with the default `n_folds=5` that is 10 fits on roughly n/10 rows each, where 64 leaves is well past what the slice supports. #1030 added `verbose=-1`, which silences the resulting "No further splits with positive gain" chatter without addressing the cause.

Two possible improvements, neither urgent:

1. A smaller `num_leaves` for this call pattern, or one sized from the data.
2. A docstring nudge on `dr_score`: comparing several candidate CATE models without an explicit learner refits 10 LGBM models per call, which is what `compute_dr_pseudo_outcomes`'s own docstring advises avoiding by precomputing pseudo-outcomes once and passing `pseudo_outcome_col`.

Measured cost when #1030's three learner-less tests were added: `tests/test_cate_scoring.py` went from 17s to 37–55s.

Observed on master (`477cd0a`).

aman-coder03:
## Proposed changes

the default learner in `compute_dr_pseudo_outcomes()`, `dr_score()`, and `plug_in_t_score()` was a fixed `LGBMRegressor(num_leaves=64, ...)`, but these functions fit per-fold and per-arm (10 fits per call with the default `n_folds=5`), each on a fraction of the data. 64 leaves is often way more than a small slice can support, which just triggers "No further splits with positive gain" warnings instead of helping.

swapped the fixed default for a sentinel that builds a learner sized to however many rows each fold/arm actually has, capped at 64 leaves like before. Also added a docstring note pointing people at `pseudo_outcome_col` when comparing multiple models, since that avoids refitting the nuisance models every time

fixes #1037 

## Types of changes

## Checklist

## Further comments

### DIFF
--- a/causalml/metrics/cate_scoring.py
+++ b/causalml/metrics/cate_scoring.py
@@ -11,6 +11,70 @@ from causalml.propensity import compute_propensity_score, compute_r_residuals
 
 logger = logging.getLogger("causalml")
 
+# Sentinel default for `learner` in compute_dr_pseudo_outcomes() / dr_score() /
+# plug_in_t_score(). These functions fit per-fold, per-arm (with the default
+# n_folds=5, that's 10 fits per call, each on a fraction of a single arm), so a
+# single fixed-size default learner is either oversized for small slices or
+# undersized for large ones. Using a sentinel instead of a pre-built estimator
+# lets _make_fold_learner() build one sized to each fit's actual row count --
+# see _sized_default_learner(). It also avoids the mutable-default-argument
+# pitfall of instantiating one shared estimator at import time.
+_AUTO = "auto"
+
+
+def _sized_default_learner(n_samples, min_child_samples=20, max_leaves=64):
+    """Build a default LGBMRegressor with num_leaves capped to what n_samples supports.
+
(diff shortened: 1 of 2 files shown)


su-jin1425:
I can see that AI was used to work on this issue. I recommend using AI to plan and reason through the problem first, and then implementing and reviewing the solution carefully. Avoid asking AI to directly solve the whole issue, as this can lead to unnecessary changes and make the PR harder to review.

Also, this PR adds quite a lot of code and documentation for a relatively small issue. Please consider simplifying the changes and keeping the PR focused on the num_leaves problem. It would also be good to add a test that directly verifies the expected num_leaves value is used for each fold, rather than only checking that the final scores are finite.

PR #1046, opened 2026-09-04
Fix #1045: seed and stratify the GBM propensity early-stopping split

### LINKED ISSUE #1045: GradientBoostedPropensityModel early-stopping split ignores random_state and is not stratified
`GradientBoostedPropensityModel.fit()` draws its early-stopping validation set with no `random_state` and no `stratify` (`propensity.py:181-184`):

```python
if self.early_stop:
    X_train, X_val, y_train, y_val = train_test_split(
        X, y, test_size=stop_val_size
    )
```

### a. `random_state` never reaches the split

`_model` (`propensity.py:154-163`) resolves the seed for the `XGBClassifier` as `self.model_kwargs.get("random_state", 42)`, so `GradientBoostedPropensityModel(random_state=42, early_stop=True)` looks fully seeded. The validation set it early-stops against is drawn from the global numpy stream instead, so every `fit()` sees a different split, stops at a different iteration, and returns a different model.

```python
import numpy as np
from causalml.propensity import GradientBoostedPropensityModel

rng = np.random.RandomState(42)
X = rng.normal(size=(1000, 10))
w = (0.8 * X[:, 0] + 0.5 * X[:, 1] + rng.normal(size=1000) > 0).astype(int)

for label, early_stop in [("early_stop=True ", True), ("early_stop=False", False)]:
    runs = []
    for _ in range(3):
        pm = GradientBoostedPropensityModel(random_state=42, early_stop=early_stop)
        runs.append(pm.fit_predict(X, w))
    same = all(np.array_equal(runs[0], r) for r in runs[1:])
    spread = max(np.abs(runs[0] - r).max() for r in runs[1:])
    print(f"{label} identical across 3 fits: {same}   max|diff| = {spread:.4f}")
```

```
early_stop=True  identical across 3 fits: False   max|diff| = 0.5466
early_stop=False identical across 3 fits: True   max|diff| = 0.0000
```

The `early_stop=False` line is the control: same seed, same data, bit-identical. That isolates the split as the only source of nondeterminism. A propensity score that moves by up to 0.55 between two identical calls propagates into every IPW/DR estimate built on it.

This is the same class of defect as #1029 (R-learner bootstraps). Notably the R-learner's *own* early-stopping split already threads the seed — `rlearner.py:822-831` passes `random_state=self.random_state` to `train_test_split`. `GradientBoostedPropensityModel` is the only early-stopping split in the library that does not.

### b. The split is not stratified on treatment

`y` here is the binary treatment indicator. Every other treatment split in the library stratifies on it:

- `compute_r_residuals` (`propensity.py:317-320`), whose docstring says "stratified on treatment so every fold retains both arms"
- `LogisticRegressionPropensityModel`'s `StratifiedKFold` (`propensity.py:113-121`)

The early-stopping split is the exception. For imbalanced treatment the validation set's treated count swings widely, and at small n it can hold no treated units at all, which makes the early-stopping metric meaningless. Over 500 random splits at `test_size=0.2`:

| n | treated rate | expected treated in val | observed min-max | splits with <2 treated |
|---|---|---|---|---|
| 400 | 10% | 8 | 2-20 | 0.0% |
| 1000 | 5% | 10 | 2-22 | 0.0% |
| 200 | 10% | 4 | 0-12 | 7.2% |

This is adjacent to #1027, which tuned the other propensity models for imbalanced treatment.

### Suggested fix

Both are the same call:

```python
X_train, X_val, y_train, y_val = train_test_split(
    X,
    y,
    test_size=stop_val_size,
    random_state=self.model_kwargs.get("random_state", 42),
    stratify=y,
)
```

Reusing the `self.model_kwargs.get("random_state", 42)` expression from `_model` keeps the split and the classifier on the same seed, and leaves `random_state=None` genuinely random per the scikit-learn convention.

Observed on master (`9a23278`).

Hari20032005:
Fixes #1045.

`GradientBoostedPropensityModel.fit()` drew its early-stopping validation set with neither a `random_state` nor `stratify`, so a model constructed with an explicit seed still produced a different fit on every call.

### The change

One call, in `causalml/propensity.py`:

```python
X_train, X_val, y_train, y_val = train_test_split(
    X,
    y,
    test_size=stop_val_size,
    random_state=self.model_kwargs.get("random_state", 42),
    stratify=y,
)
```

**`random_state`** — `_model` resolves the `XGBClassifier` seed as `self.model_kwargs.get("random_state", 42)`; reusing that expression puts the split and the classifier on the same seed. The R-learner's own early-stopping split already does this (`rlearner.py:830`), so this brings the last unseeded split in the library into line. `random_state=None` stays genuinely random, per the scikit-learn convention.

**`stratify`** — `y` here is the binary treatment indicator, and every other treatment split in the library stratifies on it (`compute_r_residuals`, whose docstring promises "every fold retains both arms"; `LogisticRegressionPropensityModel`'s `StratifiedKFold`). Without it the validation set's treated count swings by an order of magnitude on imbalanced data, and at small n can reach zero, which makes the early-stopping metric meaningless.

### Effect

Same seed, same data, three consecutive fits:

| | before | after |
|---|---|---|
| `early_stop=True` | not identical, max abs diff **0.5466** | identical |
| `early_stop=False` | identical | identical |

`early_stop=False` was already reproducible and is unchanged — it is the control that isolates the split as the source of the nondeterminism.

### Tests

Two regression tests in `tests/test_propensity.py`:

- `..._earlystopping_reproducible` — same seed gives bit-identical scores, and a *different* seed still gives different scores, so the test cannot be satisfied by hard-coding a constant.
- `..._earlystopping_stratified` — on 10%-treated data, asserts the validation split preserves the treatment rate.

Both fail on master and pass with the change. Verified the fix under numpy, pandas and polars inputs, which all produce identical scores.

### DIFF
--- a/causalml/propensity.py
+++ b/causalml/propensity.py
@@ -179,8 +179,17 @@ class GradientBoostedPropensityModel(PropensityModel):
                 so this vector must be the treatment assignment.
         """
         if self.early_stop:
+            # Seed the split with the same random_state the underlying
+            # XGBClassifier resolves to, so that early stopping -- and
+            # therefore the fitted model -- is reproducible. Stratify on the
+            # treatment indicator so the validation set retains both arms,
+            # matching how the rest of the library splits on treatment.
             X_train, X_val, y_train, y_val = train_test_split(
-                X, y, test_size=stop_val_size
+                X,
+                y,
+                test_size=stop_val_size,
+                random_state=self.model_kwargs.get("random_state", 42),
+                stratify=y,
             )
 
             self.model.fit(
--- a/tests/test_propensity.py
+++ b/tests/test_propensity.py
@@ -1,5 +1,7 @@
 import numpy as np
+import pytest
 
+from causalml import propensity
 from causalml.propensity import (
     ElasticNetPropensityModel,
     GradientBoostedPropensityModel,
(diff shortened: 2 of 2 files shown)


Hari20032005:
The Read the Docs check failed on an intersphinx network timeout, not on anything in this diff:

```
WARNING: failed to reach any of the inventories with the following issues:
intersphinx inventory 'https://docs.scipy.org/doc/scipy/objects.inv' not fetchable due to
<class 'requests.exceptions.ConnectTimeout'>: ... Connection to docs.scipy.org timed out. (connect timeout=30)
```

Since the docs build runs under `-W`, that single transient warning fails it. This PR touches only `causalml/propensity.py` and `tests/test_propensity.py` — no docs, no docstrings — so a re-run should be green.

The `build` workflows appear to be waiting on maintainer approval (first-time contributor). Locally, `tests/test_propensity.py` is 9 passed on Python 3.13 / numpy 2.5.2 / scikit-learn 1.9.0.

